PR #123, opened 2011-11-11
Bugfix: Enable connection to ASM instances

robhu:
If this check is not done it is not possible to connect to an ASM instance using the adapter, as you cannot do alter session cursor_sharing in an ASM. Trying to do causes an 'ORA-15000: command disallowed by current instance type (NativeException)' exception.


yahonda:

PR #150, opened 2012-02-16
getting raw connection from websphere wrapper

moger777:
This does not include any specs since I'm not sure how to mock out java packages. Any ideas on this?


yahonda:

PR #165, opened 2012-04-27
Bug when NUMERIC type has a SCALE of nil

CoryFoy:
This commit fixes an issue where we have a NUMERIC column with a SCALE of nil that was being considered a BigDecimal instead of an integer


yahonda:

PR #169, opened 2012-05-16
Add a :schema option to allow switching the active schema using 'alter session'

jeffperrin:
Makes a schema option available in database.yml like so:

    development:
      database: xe
      username: database_user
      password: 1234
      adapter: oracle_enhanced
      host: localhost
      schema: schema_to_use

Use case:

I work in a corporate environment that insists on running the application under a different user/schema than the schema where the application lives (ie; blah_update_user is what we connect with, but the app lives in the blah schema). We could hack around this in Rails by hard-coding schema prefixes in a `set_table_name` declaration, but this had many implications including making setup of testing databases extremely annoying.

This patch basically just runs `alter session set current_schema = schema_name` immediately after initializing a connection. We've been using a variation of this patch in several production jruby/rails3/oracle 10/11 applications for the past year without issue.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'http://rubygems.org'
 
 group :development do
-  gem 'jeweler', '~> 1.5.1'
+  gem 'jeweler'
   gem 'rspec', '~> 2.4'
   gem 'rdoc'
 
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -901,12 +901,12 @@ module ActiveRecord
 
       # Default tablespace name of current user
       def default_tablespace
-        select_value("SELECT LOWER(default_tablespace) FROM user_users WHERE username = SYS_CONTEXT('userenv', 'session_user')")
+        select_value("SELECT LOWER(default_tablespace) FROM user_users WHERE username = SYS_CONTEXT('userenv', 'current_schema')")
       end
 
       def tables(name = nil) #:nodoc:
         select_values(
-        "SELECT DECODE(table_name, UPPER(table_name), LOWER(table_name), table_name) FROM all_tables WHERE owner = SYS_CONTEXT('userenv', 'session_user') AND secondary = 'N'",
+        "SELECT DECODE(table_name, UPPER(table_name), LOWER(table_name), table_name) FROM all_tables WHERE owner = SYS_CONTEXT('userenv', 'current_schema') AND secondary = 'N'",
         name)
       end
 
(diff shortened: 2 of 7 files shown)


heilja:
It is a bit short sighted to say this request does not merit inclusion.   Organizations are being asked to increase the security of their applications.  Thus, having this feature allows developers and dba's to separate sql code deployment and application operations from schema ownership.  Having role separation is key to improving security, and allows for better audit and security compliance.

cjbj:
+1

unclebilly:
I also agree that this is a valuable feature and I would like to see it get merged.  I'm not sure it should be so heavy-handed, though... perhaps the `alter session` should not happen unless the `config[:schema]` exists?  That would at least prevent this change from breaking the adapter for people who haven't specified `schema` in their database.yml.  In other words, this behavior should be opt-in - no?  @jeffperrin

jeffperrin:
The behaviour is opt-in, see: https://github.com/rsim/oracle-enhanced/pull/169/files#diff-ca7c8c425714bbfad966883480dd2943R18

That said, this PR is 3 years old now and would need updating if it was to be mergeable. I'm no longer using Oracle at this point either, so someone else would have to do the work.

barendt:
We've been maintaining our own patch that we apply to the adapter that adds the schema option. If this concept is something that has a chance of being accepted I'd be happy to get an updated PR together.

net1957:
+1 for this
At this time we use synonyms but it's always difficult because  the DBA has more work to do, it's one more task to do and change management is more difficult.

yahonda:
I understand there are some needs for this feature and I'm thinking 

* Create another gem whose `supports_migrations?` returns `false`. Not implementing this feature to Oracle enhanced adapter itself.

* I guess many of those who needs this feature are not using Rails migration `rake db:migrate`. Then it looks difficult (for me) to implement Oracle enhanced unit tests for this feature since unit tests are working like this - creating new tables, do some tests then drop tables created.

* It overrides Oracle enhanced adapter`SYS_CONTEXT('userenv', 'session_user') ` with `SYS_CONTEXT('userenv', 'current_schema')`

barendt:
I'm not sure about the others, but in our case we do use `rake db:migrate`. We use a `RAILS_ENV` for migrations that is different from the one we use when serving up the application over the web. The reason for this is that we're using Oracle VPD to implement some additional row-level filtering at the database layer - so we run migrations against one schema and then use the VPD to populate another schema with database objects to serve up to our web users.

For what it's worth, the patch we're using to add support for a `schema` option looks like this: https://github.com/rsim/oracle-enhanced/compare/master...PMACS:schema_patch

yahonda:
@banyan Thanks for explaining your use case. My guess may be wrong.

yahonda:
#742 has been merged to master. It enables `schema` options. Thanks guys providing your use cases.

PR #304, opened 2013-04-19
Primary key additions for schema dumper

blackjk3:
We have run into a few issues trying to use the schema dumper on our legacy oracle tables that have some rails conventional tables, plus some older style composite key tables.  Also some of our legacy tables use guids which are a primary key with type string.  

I have updated the code to now support non-numeric primary keys and composite primary keys in the schema dumper.  This way we don't have to modify our schema.rb to run tests.

I have included tests for these new features in the dump spec.
- table with non-default primary key with non-numeric data type
- table with composite primary key

I would understand if this doesn't fit in with the "rails conventions" and therefore would not be helpful to be merged.  Either way, I thought it could be of use to someone else who has been working through legacy database issues.

Thanks!
Jason


yahonda:

PR #421, opened 2014-03-01
ChangeColumn type varchar->clob

gertfindel:
Allows to workaround while changing text columns to clob 
ORA-22858: invalid alteration of datatype  http://ora-22858.ora-code.com/

- Tests missing 
- Update sentence should use the functions instead of raw SQL
- Temporary column might be longer than MAX_IDENTIFIER_LENGTH


yahonda:

Closing this pull request since I do not prefer to implement this kind of sql_type changing behavior by default. I think some of them can prefer and would prefer this change, but not everyone.

In long term - Oracle 12c introduces longer varchar2 32767 byte. Although Oracle enhanced adapter has not supported it yet, I wanted to implement this feature.

PR #470, opened 2014-07-23
removing dump_schema_information method

sarsena:
Removing dump_schema_information. The latest version of ActiveRecord already calls this method at a high level.  https://github.com/rails/rails/blob/35a9f7a6353ea12dd3103366e84ccc90123c3583/activerecord/lib/active_record/railties/databases.rake#L273-L288

Having this in the oracle-enhanced gem causes duplicate insert schema migrations in the structure.sql file.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_database_tasks.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_database_tasks.rb
@@ -40,9 +40,6 @@ module ActiveRecord
         def structure_dump(filename)
           establish_connection(@config)
           File.open(filename, 'w:utf-8') { |f| f << connection.structure_dump }
-          if connection.supports_migrations?
-            File.open(filename, 'a') { |f| f << connection.dump_schema_information }
-          end
           if @config['structure_dump'] == 'db_stored_code'
              File.open(filename, 'a') { |f| f << connection.structure_dump_db_stored_code }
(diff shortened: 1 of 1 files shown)


senguidanos:
We could really use this fix. Is this PR dead? Looks like a unit test would simply assert that dump_schema_information is not called on the connection.

yahonda:
I do not reproduce this case. I'm guessing if you guys are using non-Oracle database on each RALS_ENV.

* Would you update which database and database adapter you guys are using for each `RAILS_ENV`?

`development`
`test`
`production`

* On which environment do you need this fix?

senguidanos:
We are using oracle_enhanced for all environments, each of which connects to an Oracle db. The scenario is as follows:

Assume we have a valid structure.sql to begin with, and activerecord schema_format is :sql. We create a migration and run it in local dev. Now structure.sql has two INSERT statements for the each unique schema_migration version.

We're using Rails 4.1.7

sarsena:
@senguidanos Would you mind writing tests for this? Thats all that is needed to get this merged.  I haven't had time to do so.

mxhold:
I made a separate PR (#558) that includes this change with a test (since I can't push to your repo).

yahonda:
Closed as #558 merged to master. Thanks for opening a pull request.

PR #475, opened 2014-08-08
Set sequence name automatically when new table name has more than 26 cha...

chriger:
...racters

At the moment you have to rename the sequence explicitly before you use `rename_table` if the new table name has more than 26 characters.


yahonda:
Closed as #676 has been merged to master. Thanks for opening a pull request.

PR #510, opened 2014-10-07
Improved system password prompt

wconrad:
* Try to get the password from the environment variable
  DATABASE_SYS_PASSWORD

* When reading the password from the console, do not echo it

* Cache it, so that only one prompt is needed to do a "rake db:reset"


wconrad:
The environment variable part of this was implemented in pull request #529 which is currently merged into the rails42 branch.  That's the important part of this PR, so there's no need for this one anymore.

wconrad:
I was in error.  pull request #529 was applied to master, not to the rails42 branch.

PR #521, opened 2014-11-20
exec query was not retrying after lost connection

scottschulthess:
We were having an issue on our rails 3.2 project where if the database connection was lost it would not attempt to retry because the exec_query codepath does not execute with_retry.


yahonda:
I think `select` method is already auto_retry aware.

I have tested the latest `release14` branch of Oracle enhanced adapter with Rails 3.2.22, JRuby 1.7.22.
Then these auto reconnect tests using `select` method successfully finished. 

https://github.com/rsim/oracle-enhanced/blob/release14/spec/active_record/connection_adapters/oracle_enhanced_connection_spec.rb#L260-L286

Would you have some test cases that satisfies your changes. Thanks,

scottschulthess:
Close this for now then! thanks!

yahonda:
Thanks.

PR #529, opened 2014-12-05
Read SYSTEM password from ENV ORACLE_SYSTEM_PASSWORD optionally

marcelmorgan:
This is useful when running scripts and no user is around to enter
a system password value. In particular, for various CI server setups.

Usage Example:

```
ORACLE_SYSTEM_PASSWORD=secret rake db:migrate
```


wconrad:
I was in error: The patch was applied to master, not to rails42.

PR #532, opened 2014-12-11
Show original error message when loading ruby-oci8 library fails

mcmorgan:
While I had the ruby-oci8 library already installed, I was puzzled by this error:

> ERROR: ActiveRecord oracle_enhanced adapter could not load ruby-oci8 library. Please install ruby-oci8 gem.

After revealing the original load error received, I discovered that the original message was:

> LoadError: libaio.so.1: cannot open shared object file: No such file or directory

So I got rid of this (on Travis CI), by installing `libaio1` via `sudo apt-get install libaio1`

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_oci_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_oci_connection.rb
@@ -2,9 +2,11 @@ require 'delegate'
 
 begin
   require "oci8"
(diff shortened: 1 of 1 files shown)


mcmorgan:
@yahonda I realized you missed some notifications from Github - what do you think of this PR though?

yahonda:
Merged. Thanks.

PR #541, opened 2015-01-30
Add `:if_exists` option to `drop_table` for compatibility with other adapters

kamipo:
`:if_exists` option introduced by https://github.com/rails/rails/pull/18597.
And it is used by test case already. https://github.com/rails/rails/commit/7675364

How about this pull request?


yahonda:
Thanks for the pull request.

If I understand correctly `:if_exists` has been merged to master, then this option would be available from Rails 5. Our master branch currently supports Rails 4.1 and lower version of Rails, `rails42` is a working branch to support Rails 4.2. Though I've tested this pull request and found it is harmless for Rails 4.1 or lower version of Rails, then merging it.

Thanks.

kamipo:
Thank you for merging :blush:

PR #550, opened 2015-02-06
Rails42 configurable spec

apeiros:
This patch adds the ability to load spec configuration from a file instead of env variables. This is easier to share in our team and I could imagine other scenarios where it is useful.
ENV based configuration still works too, of course.


yahonda:
Thanks.

Issue #548, opened 2015-02-04
Rails42: Deprecation warning with serialized attributes


apeiros:
The enhanced adapter adds ActiveRecord::Base#record_changed_lobs, which uses self.class.serialized_attributes. But the ActiveRecord::Base.serialized_attributes method is deprecated (see https://github.com/rails/rails/pull/15704). This generates a warning on all write operations on serialized attributes.

apeiros:
Given that http://edgeguides.rubyonrails.org/4_2_release_notes.html#active-record-notable-changes (3rd point, "ActiveRecord::Dirty now detects in-place changes to mutable values.") I wonder whether the call is superfluous as of rails 4.2. I.e., whether we could just replace:
```(self.class.serialized_attributes.keys.include?(col.name) || self.send(:"#{col.name}_changed?")) && !self.class.readonly_attributes.to_a.include?(col.name)```
with:
```self.attribute_changed?(col.name) && !self.class.readonly_attributes.to_a.include?(col.name)```

But I don't know how to implement it nicely in a backwards compatible way. Explicitly versioned implementation seems ugly.

yahonda:
Yes. `serialized_attributes` has been deprecated from Rails 4.2 and actually removed from Rails master branch https://github.com/rails/rails/commit/82043ab53cb186d59b1b3be06122861758f814b2#diff-a72ce03f1fc846ef77d68506c9060bfc

In the meantime, our primary target is to support Rails 4.2 and also there are still issues and failures to address. Then removing deprecated warning has lower priority right now.

When we start developing Oracle enhanced adapter Rails 5 support, you can implement this fix without thinking about Rails version because there are bunch of changes in ActiveRecord rails master branch I think we will need another version to support Rails 5.

apeiros:
Maybe it would make sense to change the version number to lock-step? That is, Major/Minor match rails version, rest is for your versioning. That should reduce headaches about backwards compatibility a lot. Downside is: more versions to parallely maintain.
Readying the pull requests btw., they should come in today.

sockmonk:
I'm upgrading to Rails 4.2 now, and this is a very noisy deprecation warning in my test logs; it apparently triggers every time the app saves to 'text' / lob column. I know it doesn't keep it from saving, but I hope it can be addressed before the Rails 5.x version.

yahonda:
Closed #688 should suppress this warning.

PR #552, opened 2015-02-06
Remove call to deprecated `serialized_attributes`.

apeiros:
Fixes https://github.com/rsim/oracle-enhanced/issues/548

According to http://edgeguides.rubyonrails.org/4_2_release_notes.html#active-record-notable-changes
(3rd point, "ActiveRecord::Dirty now detects in-place changes to mutable values.")
it should be no longer necessary to check whether the attribute is serialized.

Note: this change is probably incompatible with rails prior 4.2.0.


yahonda:
Merged. Thanks.

Issue #556, opened 2015-02-25
Database links are being incorrectly quoted


cdinger:
Database links are allowed to contain periods, unlike other database objects, as defined in `ActiveRecord::ConnectionAdapters::OracleEnhancedAdapter::NONQUOTED_DATABASE_LINK`. Yet database links get quoted with the `quote_column_name` method, which uses the quoting rules for columns.

This means that if I have a database link called `SOME.LINK`, a table reference across that link will get incorrectly quoted like this: `"SOME_SCHEMA"."SOMETABLE"@"SOME"."LINK"` when the quoting *should* produce `"SOME_SCHEMA"."SOMETABLE"@"SOME.LINK"`

I have a [suggested fix](https://github.com/cdinger/oracle-enhanced/commit/ed23416685367871958860714f2ca4e7ae6279a2) for this, but I'm having trouble getting the rest of the test suite to run reliably. I'm running it against an XE instance in a docker container and I'm guessing that's causing issues. What do you all use? I'm currently downloading [this](http://www.oracle.com/technetwork/database/enterprise-edition/databaseappdev-vm-161299.html) which I'm hoping works better. I'd like to run all of the test before submitting a PR.

yahonda:
Thanks for the reporting. If you have Vagrant installed box, 
https://github.com/yahonda/rails-dev-box-runs-oracle would help you to perform unit tests.

cdinger:
Thanks for the vagrant box, @yahonda. That's helpful. The issue ending up being the Rails version that it was using. I still had some failures, but they were the same failures I got from the `rails42` branch without my change.

Anyway, I've filed [a PR](https://github.com/rsim/oracle-enhanced/pull/557) for this. Thanks again!

PR #557, opened 2015-02-25
Properly quote database links

cdinger:
Database links, unlike column names, CAN contain periods, which were not being handled properly by the `quote_table_name` method (#556).

This PR:

- adds a spec demonstrating the bug
- adds a new `quote_database_link` method to be used only for database links (since link quoting follows different rules)
- reworks the existing `quote_table_name` method to differentiate between tables/schemas, and links


yahonda:
Thanks for opening the pull request. 
Unless you have some specific reason, would you open a pull request to the `master` branch?

cdinger:
@yahonda, sure. Just created #559 to master. I'll close this.

PR #558, opened 2015-02-27
Do not dump schema information during structure dump

### LINKED ISSUE #460: db:structure:dump inserts two sets of schema migration values
The resulting structure.sql file contains two sets of values for the `schema_migrations` table, resulting in an error on import due to duplicate ids.

It seems like Rails itself is calling the `dump_schema_information` method in addition to the adapter itself calling it in case the connection supports migrations, in the following from `oracle_enhanced_database_tasks.rb`:

        def structure_dump(filename)
          establish_connection(@config)
          File.open(filename, 'w:utf-8') { |f| f << connection.structure_dump }
          if connection.supports_migrations?
            File.open(filename, 'a') { |f| f << connection.dump_schema_information }
          end
          if @config['structure_dump'] == 'db_stored_code'
             File.open(filename, 'a') { |f| f << connection.structure_dump_db_stored_code }
          end
        end

mxhold:
Right now, when you run `rake db:structure:dump` the schema information (i.e. which migrations have been run) are dumped twice when using sql as the schema format. This prevents you from loading the `structure.sql` file since it will try to run inserts that violate the uniqueness constraint on the `schema_migrations` table.

This happens because Rails is calling `connection.dump_schema_information` after we also call it in `connection.structure_dump`.

This change removes our call to `connection.dump_schema_information` so it is only called once, by Rails.

This fixes #460. This is the same change as in #470 but with tests.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_database_tasks.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_database_tasks.rb
@@ -45,9 +45,6 @@ module ActiveRecord
         def structure_dump(filename)
           establish_connection(@config)
           File.open(filename, 'w:utf-8') { |f| f << connection.structure_dump }
-          if connection.supports_migrations?
-            File.open(filename, 'a') { |f| f << connection.dump_schema_information }
-          end
           if @config['structure_dump'] == 'db_stored_code'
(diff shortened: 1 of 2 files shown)


sarsena:
:+1:

yahonda:
Merged. Thanks for the pull request.

PR #587, opened 2015-05-01
Fixed typo in the rake tasks load statement

tzumby:
There was a typo in the load statement for the database tasks, fixed the load path.

rake aborted!
LoadError: cannot load such file -- active_record/connection_adapters/oracle_enhanced_database_tasks.rb


yahonda:
Merged. Thanks for testing and the pull request.

PR #598, opened 2015-05-13
Use Oracle BINARY_FLOAT datatype for Rails :float type

yahonda:
This pull request changes mapping for Rails :float type from Oracle NUMBER to BINARY_FLOAT.

* Based on this document `BINARY_FLOAT` is supported from 10.1 (aka 10gR1).
http://docs.oracle.com/cd/B14117_01/server.101/b10759/wnsql.htm#sthref11
9iR2 or lower version cannot work but it should not be a problem since it was released on 2003.

* binary_float or binary_double

Based on Rails 4.2 implementation, `binary` is configured as a alias for `float`. Then it should be enough to support `BINARY_FLOAT` which should be corresponding to `float` in other databases.
https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L390-L419

* Compatibility
This pull request is for rails42 branch only. It should be released as Oracle enhanced adapter 1.6 to support Rails 4.2. I do not think to backport it Oracle enhanced adapter 1.5.0 since
it changes public behavior. As of right now :float is mapped to Oracle NUMBER datatype.

Also this pull request addresses following failure.

```ruby
$ ARCONN=oracle ruby -Itest test/cases/attributes_test.rb -n test_overloaded_properties_save
Using oracle
Run options: -n test_overloaded_properties_save --seed 53381

# Running:

F

Finished in 0.235332s, 4.2493 runs/s, 16.9972 assertions/s.

  1) Failure:
ActiveRecord::CustomPropertiesTest#test_overloaded_properties_save [test/cases/attributes_test.rb:43]:
Expected #<BigDecimal:1ee3300,'0.2E1',9(27)> to be a kind of Float, not BigDecimal.

1 runs, 4 assertions, 1 failures, 0 errors, 0 skips
$
```

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -459,7 +459,7 @@ module ActiveRecord
         :string      => { :name => "VARCHAR2", :limit => 255 },
         :text        => { :name => "CLOB" },
         :integer     => { :name => "NUMBER", :limit => 38 },
-        :float       => { :name => "NUMBER" },
+        :float       => { :name => "BINARY_FLOAT" },
         :decimal     => { :name => "DECIMAL" },
         :datetime    => { :name => "DATE" },
         # changed to native TIMESTAMP type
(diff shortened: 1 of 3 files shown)


yahonda:
@rsim Would you review this pull request since it changes public/external behavior.

PR #642, opened 2015-06-25
Prevent Rails 4.2 using version 15 series.

Eric-Guo:
It's may save some developer time like [ruby-china discussed](https://ruby-china.org/topics/26161).


yahonda:
Would you squash these two commits?

Eric-Guo:
:squirrel: done

yahonda:
Thanks.

PR #668, opened 2015-07-14
create_table should use default tablespace values for lobs

substars:
Applying  #667 to master.

I couldn't do a bundle install using rails trunk:
```
└─╼ bundle update
Updating git://github.com/rails/rails.git
Updating git://github.com/rails/arel.git
Updating git://github.com/rails/journey.git
Updating git://github.com/kubo/ruby-oci8.git
Fetching gem metadata from http://rubygems.org/.........
Fetching version metadata from http://rubygems.org/...
Fetching dependency metadata from http://rubygems.org/..
Resolving dependencies...
Bundler could not find compatible versions for gem "arel":
  In Gemfile:
    arel (= 7.0.0.alpha) java

    activerecord (>= 0) ruby depends on
      arel (= 7.0.0.alpha) ruby

    arel (>= 0) java

    arel (>= 0) ruby
```

I ran tests against 4-2-stable:
```
  gem 'activerecord',   github: 'rails/rails', branch: '4-2-stable'
  gem 'activemodel',    github: 'rails/rails', branch: '4-2-stable'
  gem 'activesupport',  github: 'rails/rails', branch: '4-2-stable'
  gem 'actionpack',     github: 'rails/rails', branch: '4-2-stable'
  gem 'railties',       github: 'rails/rails', branch: '4-2-stable'

  gem 'arel',           github: 'rails/arel', branch: '6-0-stable'
```

Is this a problem for everyone or just me?

PR #670, opened 2015-07-16
create_table should use default tablespace values for lobs

substars:
backport #668 to 1.5 branch since we still have Rails 4.1 apps :older_man: :older_woman:


yahonda:
Thanks.

PR #672, opened 2015-07-20
Requires activerecord 4.2.1 or higher, not activemodel

yahonda:
This pull request follows up #652 since it added incorrect dependency. It requires activerecord, not activemodel.


Eric-Guo:
Can you explain why need activemodel instead of activerecord here? Or simply because activerecord depend on activemodel? in my understanding, acitverecord provide all relative database function instead of activemodel.

Anyway just curious.

yahonda:
Actually I do not know. Since I'm having the same question. Wanted to`Gemfile` and `activerecord-oracle_enhanced-adapter.gemspec`, which were relatively lower priority. Thinking about to do it when we support Rails 5. Pull requests are welcomed.

Issue #678, opened 2015-08-03
AnyModel.offset(n) causes undefined method `expr' for n:Fixnum


apeiros:
With rails 4.2.3 and Oracle Enhanced Adapter 1.6.2 I experience the following problem:

```
class MyModel < ActiveRecord::Base; end
MyModel.offset(0).first # !> undefined method `expr' for 1:Fixnum
```

I wrote tests for it and will submit a pull request to add the failing tests.
I'll also try to figure what's going wrong and how to fix it, maybe I'm lucky and can figure it out.

apeiros:
Great, the problem seems to be with Arel and is solved in master. Specifically this commit: https://github.com/rails/arel/commit/74563c63bdda08fdf1a026e908d51f536436c674
I suggest to just pull in the PR to have the specs and then close the issue.

apeiros:
For those who run into the same issue - until the fix is rolled into a proper release, this is what you can do: Add `gem 'arel', github: 'rails/arel', branch: '6-0-stable'` to your Gemfile. This resolves the bug wrt AR::Base.offset for me.

yahonda:
Thanks for reporting. 
Arel 6.0.3 has been released with https://github.com/rails/arel/pull/377 fix included. Closing this issue.

PR #679, opened 2015-08-03
Add minimal specs for ActiveRecord::Base.limit() and .order().

apeiros:
Specs for #678


yahonda:
Merged. Thanks for adding tests.

PR #685, opened 2015-08-11
first pass of changes to support pmacs system

walterdavis:
There is still one chunk missing, as noted in an e-mail to all of you recently. If you can help me figure out where that goes, I can finish this and we can begin adapting the tests to see if it still works.


walterdavis:
pastedgraphic-1
As noted elsewhere, I am lost figuring out where this code should live in the new gem structure.

PR #709, opened 2015-09-04
Revert "Update the gemfile entry so that users can use version 1.6.3 …

yahonda:
…or higher"

This reverts commit cf08dbeb856d59be48ba39db045ec1429b6f061d.


yahonda:
It was my bad.

PR #727, opened 2015-10-05
Update dirty.rb

jadon1979:
Fix issue with _field_changed? "wrong number of args 2..3" error that keeps popping up.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/dirty.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/dirty.rb
@@ -5,7 +5,9 @@ module ActiveRecord #:nodoc:
       module InstanceMethods #:nodoc:
         private
 
-        def _field_changed?(attr, old_value)
+      
+        def _field_changed?(*args)
+          attr, old_value = args
           new_value = read_attribute(attr)
           raw_value = read_attribute_before_type_cast(attr)
 


yahonda:
Thanks for opening a pull request. would you update which version of ActiveRecord you are using.

jadon1979:
activerecord (4.2.4)

yahonda:
Thanks for the update. I have checked the method signature of parent method in ActiveRecord 4.2.4. which has not changed.

https://github.com/rails/rails/blob/v4.2.4/activerecord/lib/active_record/attribute_methods/dirty.rb#L141-L143

```ruby
      def _field_changed?(attr, old_value)
        @attributes[attr].changed_from?(old_value)
      end
```

Would you provide a test case to reproduce "wrong number of args 2..3" error.
Thanks.

jadon1979:
As soon as I updated to 1.5 any record.update & create would throw the error in both console and in app.  I don't have the time right now to provide more than that.  I can update this later.

yahonda:
Oracle enhanced adapter 1.5.x does not support ActiveRecord 4.2. Then would you confirm that you are using Oracle enhanced adapter 1.6.3.

```ruby
gem 'activerecord-oracle_enhanced-adapter', '~> 1.6.0'
```
Still you can reproduce this error by using Oracle enhanced adapter 1.6.3 please update the test case to reproduce.

jadon1979:
That worked.  Thanks.  Closing this now.

yahonda:
Thanks.

PR #728, opened 2015-10-06
Add note to readme about adapter name when using DATABASE_URL

cdinger:
@yahonda I anticipate this will come up more and more. Would you consider adding a note about it in the README?


yahonda:
Thanks.

Issue #660, opened 2015-07-01
dead connections


frankapimenta:
I have a rails application that connects to postgres and oracle databases thru different models.
Regarding the conn to the oracle database, the firewall kills connections that are not being used after an amount of time (maybe 60m). When we try to reconnect to the oracle db the call hangs for nearly 15m to work then again. 

Is there a way that the gem covers this -> http://www.calebwoods.com/2014/10/07/oracle-connection-hanging-ruby/

I checked line 706 in https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb

```ruby
code -- 
      # If SQL statement fails due to lost connection then reconnect
      # and retry SQL statement if autocommit mode is enabled.
      # By default this functionality is disabled.
      attr_reader :auto_retry #:nodoc:
      @auto_retry = false

      def auto_retry=(value) #:nodoc:
        @auto_retry = value
        @connection.auto_retry = value if @connection
      end
-- code
```
aren't this code misleading in class instance variables and instance variables (@auto_retry)?

yahonda:
There is almost nothing that Oracle enhanced adapter gem prevents from firewall kill 'idle' sessions. I refereed the link and you can contact your system administrator and/or DBA to configure `SQLNET.EXPIRE_TIME` at database server. 

Since Oracle enhanced adapter relies on Oracle Network Services (aka SQL*Net) to connect to database. Oracle Network Services also relies on TCP/IP in most cases. Killing connections are done at this TCP/IP layer, this is out of control from Oracle enhanced adapter.

yahonda:
Closing this issue. If you have some more questions about firewall killing existing Oracle connections, you can contact other Q&A site or other support services.

cjbj:
@frankapimenta You could try tuning some of the timeout parameters in sqlnet.ora and perhaps look at the OS keep_alive setting.

Other "high availability" features like Oracle's FAN support would need changes to ruby-oci8.

Issue #420, opened 2014-02-26
Reaper not detecting closed connections


lightswitch05:
Database connections die after an hour of inactivity - reaper doesn't detect the closed connections. After waiting on the dead connection for 15 minutes, rails throws an error and rebuilds the connection. The second request works as expected. Issue exists for both production & development.

I'm using an Oracle Enhanced (v1.5.2) JDBC connetion with Rails v4.1.0.beta1 running on jBoss 7 with jRuby 1.7.10 connecting to Oracle 11g Release 2 (11.2.0.4).
## database.yml

```
default: &default
  adapter: oracle_enhanced
  database: ****
  host: ****
  port: 1521
  username: ****
  password: ****
  reaping_frequency: 150 #seconds = 2.5 minutes

development:
  <<: *default
  pool: 5

production:
  <<: *default
  pool: 10
```
## Error

```
07:14:55,068 INFO  [org.apache.catalina.core.ContainerBase.[jboss.web].[default-host].[/pricing_management]] (http--0.0.0.0-8280-1) 
Java::JavaSql::SQLRecoverableException (Closed Connection):
    oracle.jdbc.driver.PhysicalConnection.needLine(PhysicalConnection.java:6295)
    ....
    rubyjit.ActiveRecord::ConnectionAdapters::OracleEnhancedJDBCConnection::Cursor$$close_92C646BEB33B544DAAAC963FA7290D6A33D8EA591197350755.__file__(gems/gems/activerecord-oracle_enhanced-adapter-1.5.2/lib/active_record/connection_adapters/oracle_enhanced_jdbc_connection.rb:421)
```

[server.log](https://gist.github.com/lightswitch05/9231511#file-server-log), [Gemfile](https://gist.github.com/lightswitch05/9231511#file-gemfile), [Gemfile.lock](https://gist.github.com/lightswitch05/9231511#file-gemfile-lock), [Database.yml](https://gist.github.com/lightswitch05/9231511#file-database-yml)

frankapimenta:
maybe this? http://www.calebwoods.com/2014/10/07/oracle-connection-hanging-ruby/

PR #730, opened 2015-10-15
DISCUSS: Fix auto_retry on Cursors and add timeouts support

### LINKED ISSUE #419: Application hangs up after Oracle restart
When Oracle db restart, my application hangs up. Cant find where I can set any timeout parameter or auto reconnect or sth like this. Is this adapter supports any?

vjt:
Hello ORACLE's friends (pun intended) :smile: 

This quickly cooked set of commits adds support for the read/send timeout (undocumented) feature exposed by the [OCI API](http://blog.jiubao.org/2015/01/undocumented-oci-handle-attributes.html) and available on recent (> 11.1) ORACLEs.

These changes fixes #419 and #660, avoids (AFAICT) the need for a reaper (#420) and as well provides a cleaner way of dealing with dangling connections than kubo/ruby-oci8#56.

What `.recv_timeout` and `.send_timeout` do is to call `setsockopt()` on the ORACLE network socket, setting a read or write timeout to the number of *seconds* provided. When a `read()` or `write()` are idle for more than the specified seconds, they return an error to the caller that then OCI8 yields to us as an `OCI8Error` and gets fed up the chain as an `ActiveRecord::StatementInvalid`.

So, no more stuck connections waiting on the default 2 hours TCP keepalive timeout. Kittens rejoice :smile_cat:!

However, when a timeout is detected, the OCI libraries close the underlying connection, resetting the configured timeouts. Moreover, the `auto_retry` code is currently faulty on newer ARs. The `auto_retry` functionality wraps the OCI8 connection and it works when `.exec` is used on it, but it is bypassed when cursors are used. This scenario is very common in recent ARs, because the newer `exec_with_returning` AR APIs uses Cursors.

So, I have wrapped the most I could the `auto_retry` logic bits in reusable methods, and I've added `auto_retry` support to the Cursor class and **manually** tested.

These changes only targets the OCI connection and cursors for now, and it is only manually tested. But given it fixes many long standing issues, I think it is worth for you to review the changes so we can then together proceed in further refining and finalizing them.

Thanks for your attention!

~Marcello


yahonda:
Thanks for opening a pull request. Let me take a look at this.

vjt:
Other than send/receive timeouts, I'd add that another timeout setting is relevant to this issue, [CONNECT_TIMEOUT](http://docs.oracle.com/cd/E11882_01/network.112/e10835/tnsnames.htm#NETRF666). This unfortunately can't be configured through OCI API, and it must be configured through the connect string.

A full `database.yml` example displaying the three timeouts configuration follows, and it is what we're currently using in staging.

```yaml
staging:
  database: "(DESCRIPTION=
    (ADDRESS=(PROTOCOL=TCP)(HOST=oracle.example.com)(PORT=1521))
    (CONNECT_TIMEOUT=5)(TCP_CONNECT_TIMEOUT=5)
    (CONNECT_DATA=(SERVER=DEDICATED)(SERVICE_NAME=DATABASE_SERVICE)))"
  username: database_user
  password: database_password
  adapter: oracle_enhanced
  recv_timeout: 10
  send_timeout: 2
```

yahonda:

Would you consider to divide this pull request so that (at least I) can understand each pull request more clearly.

* Setting `recv_timeout` and `send_timeout` via OCI
https://github.com/ifad/oracle-enhanced/commit/9470c272746af3444c0e3aaf1bbc3192e0053727

As you already pointed out OCI_ATTR_RECEIVE_TIMEOUT and OCI_ATTR_SEND_TIMEOUT has been implemented in 11.1. They are not documented  yet in 12.1 manuals. 
Also ruby-oci 2.1.8 or higher version required.

According to https://github.com/kubo/ruby-oci8/commit/b3406438089fe83d54aee1ad90bc60c99f4bf965
There are sqlnet.ora parameters sqlnet.ora SQLNET.SEND_TIMEOUT and SQLNET.RECV_TIMEOUT parameters, which should work same as OCI_ATTR_RECEIVE_TIMEOUT and OCI_ATTR_SEND_TIMEOUT parameters. 
They have been implemented in 10.1 and also documented.
No dependencies with ruby-oci8 version.

Then we have an option for users using older version of Oracle/ruby-oci8 to use sqlnet.ora profile to have some made effects. 

* auto_retry support for cursors

I think this is the most important part of this pull request. Have not reviewed in detail yet, but wanted to have some tests to support this changes.  As of right now `auto_retry` method is tested like this: killing current session by executing `alter system kill session` statement. 

https://github.com/rsim/oracle-enhanced/blob/master/spec/active_record/connection_adapters/oracle_enhanced_connection_spec.rb#L242-L287

Would you tell me how do you test manually so that we can consider how to implement in unit tests.

* Other changes
https://github.com/ifad/oracle-enhanced/commit/386a860dfc3b1b45436e4778884b888b99f91bd6

Again, thanks for the detailed investigation and pull request.

westlakem:
Was this ever merged in?  I'm working with a legacy system and would like to limit sql statements to 5 seconds, since we have some sql that runs for hours (not kidding)

vjt:
Hi @westkalem,

not merged yet, but as you can see in the comments, statement timeout can be configured using TNSNAMES options. I can't guarantee on the behaviour, though, as I haven't that approach. If yiu can and report back, that would be helpful.

Cheers,

~Marcello
-- 
~ vjt@openssl.it
~ http://sindro.me/

vjt:
@yahonda after some tests, I agree with you that receive and send timeouts should be set through the sqlnet.ora parameters: all connection settings are in one place, more compatible with ORACLE tools, and compatible with older ORACLE versions.

I will then split this pull and rerrange/squash the commits where appropriate.

Thanks!

Issue #259, opened 2012-11-29
Table and column comments do not appear in schema or structure dump.


juanitogan:
Any hope of getting table and column comments in the schema and/or structure dumps soon?  Or did I miss something?  I can add comments to the migration (thank you to oracle_enhanced) but lose them with the dumps.  Without this feature I will have to stop using migrations and will have to use something else to manage the data model.  Table and column comments are an extremely important part of our documentation.

kodram:
:+1:

AgeevAndrew:
:+1:

yahonda:
#734 has been merged to master.

PR #734, opened 2015-10-30
Include table/column comments in schema and structure dumps

cdinger:
This PR addresses #259. I don't personally use comments, but it makes sense that if they're accepted in migrations that they should also appear in the dumps. Thoughts?


yahonda:
Thank you!

PR #741, opened 2015-11-20
Fix issue with schema dumper dumping decimal (number) datatypes with …

jprewitt:
…no precision specified correctly

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -1052,7 +1052,9 @@ module ActiveRecord
         end.map do |row|
           limit, scale = row['limit'], row['scale']
           if limit || scale
-            row['sql_type'] += "(#{(limit || 38).to_i}" + ((scale = scale.to_i) > 0 ? ",#{scale})" : ")")
+            unless row['sql_type'] == 'NUMBER' && limit.nil?
+              row['sql_type'] += "(#{(limit || 38).to_i}" + ((scale = scale.to_i) > 0 ? ",#{scale})" : ")")
+            end
           end
 
           if row['sql_type_owner']
--- a/spec/active_record/connection_adapters/oracle_enhanced_schema_dump_spec.rb
+++ b/spec/active_record/connection_adapters/oracle_enhanced_schema_dump_spec.rb
@@ -454,6 +454,31 @@ describe "OracleEnhancedAdapter schema dump" do
       end
     end
   end
+  
+  describe ":decimal datatype" do
+    before(:each) do
+      schema_define do
+        create_table :test_decimals, force: true do |t|
+          t.decimal :hourly_rate
+          t.decimal :tax_rate, precision: 10, scale: 5
+        end
+      end
+    end
+
+    after(:each) do
+      schema_define do
+        drop_table :test_decimals
+      end
+    end
+
+    it "should dump decimal type with no specified precision correctly" do
+      standard_dump.should =~ /t\.decimal "hourly_rate"$/
+    end
+    
+    it "should dump decimal type with specified precision correctly" do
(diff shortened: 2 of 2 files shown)


jprewitt:
This fixes an issue where a table migration with a column specified with a decimal datatype without precision gets schema dumped as an integer datatype with a precision of 38 and scale of 0.

### migration_file.rb
```ruby
create_table :test do |t|
  t.decimal :test_column
end
```

gets created as NUMBER (No precision or scale) in the database, which is correct.  When the schema is dumped, however, you get back:

### schema.rb
```ruby
create_table :test do |t|
  t.integer :test_column, precision: 38, scale: 0
end
```

which is incorrect.  This pull request fixes this issue.

yahonda:
Thanks for opening a pull request. There is a regression caused by this commit.

```
$ ARCONN=oracle ruby -Itest test/cases/defaults_test.rb -n test_default_integers
Using oracle
Run options: -n test_default_integers --seed 603

# Running:

F

Finished in 0.068433s, 14.6127 runs/s, 14.6127 assertions/s.

  1) Failure:
DefaultTest#test_default_integers [test/cases/defaults_test.rb:24]:
Expected #<BigDecimal:55bacc216a88,'0.1E1',9(18)> to be an instance of Fixnum, not BigDecimal.

1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
$
```
This regression should be addressed to merge this pull request.

by the way as you may be aware of this Oracle `DECIMAL` and `INTEGER` are alias for NUMBER datatype. In Rails 4.2 there is a change how to map Oracle datatype with Rails datatype. 
It used to start with Rails datatype then each decimal and integer datatype can be mapped to Oracle datatype, but in Rails 4.2 it starts with Database data type NUMBER. Then it is difficult to map Oracle NUMBER datatype with Rails integer or decimal type. Then there is a kind of workaround implemented.

https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L1277-L1281

The code above may be related with your issue.

yahonda:
In Oracle decimal without precision/scale and integer will be created as NUMBER(38).
From Oracle NUMBER datatype, it has gone it was defined as decimal or integer.

```sql
SQL> 
  1  create table decimal_integer (
  2  dec1 decimal,
  3  int1 integer
  4* )

SQL> desc decimal_integer
 Name                                      Null?    Type
 ----------------------------------------- -------- ----------------------------
 DEC1                                               NUMBER(38)
 INT1                                               NUMBER(38)

SQL>
```

Then Oracle enhanced adapter handles Oracle number data type with scale 0 as Rails integer. else it will be decimal.
https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L1273-L1282

To my better understanding would you tell me if is there any reason you need to use integer datatype without precision/scale.

jprewitt:
We wanted the NUMBER datatype (without precision) so we could store arbitrary decimal numbers as is without getting rounded or truncated by Oracle.  Oracle says this is possible in http://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#i16209 
```
For numeric columns, you can specify the column as:

column_name NUMBER 
Optionally, you can also specify a precision (total number of digits) and scale (number of digits to the right of the decimal point):

column_name NUMBER (precision, scale) 
*If a precision is not specified, the column stores values as given*. If no scale is specified, the scale is zero.
```
There is no rails number column data type option in oracle_enhanced, and I thought decimal was an alias for a NUMBER type without precision. (I was wrong on this account).

In the version of enhanced_adapter we are working with 1.5.5 Inoticed the float column type translates to an Oracle NUMBER type, but would then be dumped back into schema.rb as rails decimal type.

I guess what we want is a way to specify a NUMBER datatype in a migration without precision specified that defaults to the semantics mentioned earlier of storing the number as given.

It looks like this is the wrong fix for this issue, so i'll retract the pull request.

Thanks for your help!

PR #751, opened 2015-12-11
Set default value of Booleans correctly

### LINKED ISSUE #750: Booleans with a default value of 'N' are not instantiated correctly
My test is in this gist

https://gist.github.com/IanWhitney/cca22c4ab8ff1173ec1a#file-oracle_enhanced_data_types_spec-rb-L390

I'm adding some tests to the oracle_enhanced_data_types_spec.rb file to exercise how columns with Y/N defaults are handled when creating new records.

[This test](https://gist.github.com/IanWhitney/cca22c4ab8ff1173ec1a#file-oracle_enhanced_data_types_spec-rb-L399) fails.

```
1) OracleEnhancedAdapter boolean type detection based on string column types and names default values in new records when emulate_booleans_from_strings is false are Y or N
     Failure/Error: expect(subject.manager_yn).to eq('N')

       expected: "N"
            got: "f"

       (compared using ==)
```

I'm getting "f" because of [this line](https://github.com/rsim/oracle-enhanced/blob/37f46d4c0a788d831814b57ce4eaa41ff0505535/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L1078)

```
row['data_default'] = false if row['data_default'] == "N"
```

Which does not check the `OracleEnhancedAdapter.emulate_integers_by_column_name` config.

I have a fix for this in [my team's fork](https://github.com/umn-asr/oracle-enhanced/blob/fix_default_boolean/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L1078) which I am going to submit as a PR.

IanWhitney:
Fixes #750 

This is not the prettiest fix, but it does fix the problem. And it adds tests to catch the various cases of default values and the emulate_booleans_from_strings setting.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -1075,7 +1075,7 @@ module ActiveRecord
             row['data_default'].sub!(/^'(.*)'$/m, '\1')
             row['data_default'] = nil if row['data_default'] =~ /^(null|empty_[bc]lob\(\))$/i
             # TODO: Needs better fix to fallback "N" to false
-            row['data_default'] = false if row['data_default'] == "N"
+            row['data_default'] = false if (row['data_default'] == "N" && OracleEnhancedAdapter.emulate_booleans_from_strings)
           end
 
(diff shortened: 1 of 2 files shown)


yahonda:
Thanks!

PR #754, opened 2015-12-24
ActiveRecord::Calculations#count no longer accepts an options hash argument

riyengar8:
My apologies for such a tiny pull request. I am new to this project, and I'm just trying to dip my toe in to make sure I have the correct protocol for contributions. Please let me know if I have done something incorrectly here.

This pull request fixes a single spec that started failing in the rails 5 branch because it used some deprecated ActiveRecord syntax. Thanks.


yahonda:
Thanks for the pull request. No need to worry that a pull request is tiny as long as it addresses the problem or adds some features. The commit message shows what it addresses. Great job.

ref: https://github.com/rails/rails/commit/78dab2a8569408658542e462a957ea5a35aa4679

PR #756, opened 2016-01-05
Clean up specs and test documentation

substars:
The specs had deprecation warnings and dead code in a few places, so I resolved them. It looks like RAILS_GEM_VERSION isn't needed anymore, so I removed that and updated RUNNING_TESTS to remove that and include info on YAML config.

One question I have is which Ruby versions 1.6.x officially supports. RUNNING_TESTS.md was way out of date and I needed to edit it anyway. so I took a guess but might be wrong, e.g. MRI 1.9.3 compatibility might still be good to test, but probably not 1.8.7 :smile: 

WDYT?

### DIFF
--- a/RUNNING_TESTS.md
+++ b/RUNNING_TESTS.md
@@ -16,7 +16,9 @@ to increase process limit and then restart the database (this will be necessary 
 Ruby versions
 -------------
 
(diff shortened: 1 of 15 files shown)


yahonda:

We do not define "official" supported version (yet). 
`MRI 2.1.x and 2.2.x, and JRuby 1.7.x and 9.0.x.x.` are good candidate since I have completed Oracle enhanced unit tests and ActiveRecord unit tests using these MRI(CRuby) and JRuby versions - MRI 2.1.8 and 2.2.4 and JRuby 1.7.23 and 9.0.4.0.

Issue #755, opened 2016-01-04
An instance of TimeWithZone class does NOT quote `TO_DATE' function on Rails 4.2.x


koic:
Is this intentionally changed?

## rails 4.1.13 / activerecord-oracle_enhanced-adapter (1.5.6) (Expected)

Call ActiveRecord::Base.where method with an instance of TimeWithZone argument.

```ruby
> created_at = Model.first.created_at
> Model.where(created_at: created_at)
  Model Load (5.1ms)  SELECT "MODELS".* FROM "MODELS"  WHERE "MODELS"."CREATED_AT" = TO_DATE('2015-12-28 03:55:04','YYYY-MM-DD HH24:M
I:SS')
=> [#<Model id: ***snip***>]
```

Returning an Array of matched instances, as I had expected.

## rails 4.2.5 / activerecord-oracle_enhanced-adapter (1.6.5) (Actual)

Ditto. Call ActiveRecord::Base.where method with an instance of TimeWithZone argument.

```ruby
> created_at = Model.first.created_at
> Model.where(created_at: created_at)
  Model Load (11.8ms)  SELECT "MODELS".* FROM "MODELS" WHERE "MODELS"."CREATED_AT" = :a1  [["created_at", 2015-12-28 03:55:04 UTC]]
=> []
```

This SQL does NOT quoted `TO_DATE` function. Then returning an empty Array. (ummm)

If that is NOT intentional, then patch is difficult to me...

yahonda:
Thanks for reporting. Would you create a minimum and entire test case to reproduce your issue.

```ruby
# Activate the gem you are reporting the issue against.
gem 'activerecord', '4.2.5'
gem 'activerecord-oracle_enhanced-adapter', '1.6.5'
require 'active_record'
require 'minitest/autorun'
require 'logger'
require 'active_record/connection_adapters/oracle_enhanced_adapter'

# Ensure backward compatibility with Minitest 4
Minitest::Test = MiniTest::Unit::TestCase unless defined?(Minitest::Test)

# Set Oracle enhanced adapter specific connection parameters
DATABASE_NAME = ENV['DATABASE_NAME'] || 'orcl'
DATABASE_HOST = ENV['DATABASE_HOST']
DATABASE_PORT = ENV['DATABASE_PORT']
DATABASE_USER = ENV['DATABASE_USER'] || 'oracle_enhanced'
[...]

PR #778, opened 2016-02-12
[WIP] use JDBC implicit statement cache instead of AR StatementPool

substars:
Context: we've noticed a few issues with the statement caching implementation in our JRuby apps:
* `statement_limit: 0` doesn't work, raises NoMethodErrors trying to delete from an empty hash.  You can specify `statement_limit: 1` to effectively disable statement caching, but statements stay open longer than they really should and you also have unnecessary hash operations going on.
* We wanted to disable statement caching in the first place because we would get Protocol Violation errors from our RAC because of cached PreparedStatements. This can happen if you change a table and then use a cached `SELECT *` PreparedStatement [1], which is not unusual for ActiveRecord.
* ojdbc has a built-in statement cache that is intended to be enabled, has lots of documentation, etc.  [2].

The Oracle cache seems to recover from errors more easily; the above problem causes Protocol Violation errors from when we run a migration until we restart Rails, which is pretty bad. The implicit cache doesn't completely solve the problem, but this PR also moves the caching part into `with_retry`, so we could clear the cache and retry if needed.

So what I'm proposing here is to:
* Make oracle's implicit cache the default for JRuby.
* Make the cache implementation swappable so users could implement their own cursor (for e.g. explicit caching or some other strategy).

There are still a few failing tests since ojdbc doesn't expose the same information as StatementPool, will fix those is there are no other blockers here. Thoughts?

[1] We thought this was a bug, but Oracle has a support doc ID (1361340) if you can access their support site. We see it in our RAC but can't reproduce it in an XE VM.
[2] "Oracle JDBC drivers are designed with the supposition that implicit caching is enabled" - http://www.oracle.com/technetwork/articles/vasiliev-oracle-jdbc-090470.html

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
@@ -2,6 +2,7 @@ module ActiveRecord
   module ConnectionAdapters
     # interface independent methods
     class OracleEnhancedConnection #:nodoc:
+      DEFAULT_STATEMENT_LIMIT = 250
 
       def self.create(config)
         case ORACLE_ENHANCED_CONNECTION
--- a/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb
@@ -12,7 +12,7 @@ module ActiveRecord
         end
 
         def clear_cache!
-          @statements.clear
+          @connection.clear_cache!
           reload_type_map
         end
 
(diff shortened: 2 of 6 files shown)


yahonda:

I've read some documents about implicit statement caching features, including Oracle support note 1361340.1.

I have a question about this pull request.

Do you see these exceptions occur when using Rails-native StatementPool implementation, I mean current implementation of Oracle enhanced adapter.

If I understand it correctly, Oracle enhanced adapter does not use Oracle implicit statement caching (then you have opened this pull rquest) then one of solution: `2. do not use Statement Caching` is already satisfied with current implementation of Oracle enhanced adapter.

Or does `Statement Caching` mean more generic statement caching?

substars:
Thanks for looking intio this! I think I should maybe have put the Oracle implicit caching into a different PR because it confuses the issue here.

> Do you see these exceptions occur when using Rails-native StatementPool implementation, I mean current implementation of Oracle enhanced adapter.

Yes. The support doc points out that "this limitation also applies to when explicitly reusing Statements in the code," which is what the adapter is doing by having its own cache. The query fails with the same error until we restart the app. I don't think `auto_retry` can be made to work the way the code is now because the cache lookup happens outside the `with_retry` block, so you can't reconnect and also evict the invalid cached statement.

Setting `statement_limit: 1` is a workaround so Cursor objects are never reused, but seems very inefficient.

So I would like to: 
* Make the statement caching strategy easier to disable (see https://github.com/rails/rails/issues/23075 for one idea, looks like oracle-enhanced does not support a `prepared_statements` setting).
* Move the StatementPool implementation to `jdbc_connection` and `oci_connection`. Disabling it entirely causes performance degradation, we probably need to reset the connection like `auto_retry` does and also clear the statement cache (this would not be a problem using Oracle implicit cache because the cache resets along with the connection).
* Add `Protocol Violation` to the list of retryable errors.

This means we could try disabling the cache or retrying the query with a new connection (after clearing the cache).

Additionally, I think the following is a good idea, though maybe a separate PR:
* Use Oracle's implicit statement cache by default for JRuby.

This moves Ruby code into Java and follows Oracle recommendations. I know that the support doc also mentions this caching, but it is still an improvement (besides moving Ruby into faster Java and following Oracle's recommendations) because the cache lives and dies with the connection instead of needing to be managed manually. But like I said above, we could do this ourselves by subclassing if oracle-enhanced implemented the first part.

yahonda:
Thanks for the update. Let me update about `prepared_statements` setting.

> Make the statement caching strategy easier to disable (see rails/rails#23075 for one idea, looks like oracle-enhanced does not support a prepared_statements setting).

`prepared_statements => false` works at Oracle enhanced adapter at least current master branch. I tested like this #755, have not tested using database.yml though.
```ruby
CONNECTION_PARAMS = {
  :adapter => "oracle_enhanced",
  :database => DATABASE_NAME,
  :host => DATABASE_HOST,
  :port => DATABASE_PORT,
  :username => DATABASE_USER,
  :password => DATABASE_PASSWORD,
  :prepared_statements => false
}
```

Will update and reply other suggestions / proposals.

yahonda:
Would you open another pull requests for these changes. At this moment, current `StatementPool` implementation should work both for oci and jdbc connections. 

No code change necessary to support `prepared_statements` configuration, as I said.

> * Move the StatementPool implementation to jdbc_connection and oci_connection. Disabling it entirely causes performance degradation, we probably need to reset the connection like auto_retry does and also clear the statement cache (this would not be a problem using Oracle implicit cache because the cache resets along with the connection).
> * Add Protocol Violation to the list of retryable errors.

yahonda:
Update: As long as each change has its own commit, it is also fine.

substars:
Got it, will put together another PR soon :fireworks:

PR #783, opened 2016-02-25
Support t.foreign_key use the same `to_table` twice

yahonda:
refer rails/rails#23614


Eric-Guo:
Since the Rails 4.2.6 is released, so urgent need release this change because otherwise it will break the db:migrate.

yahonda:
Thanks for the update.

yahonda:
Will release the newer version of Oracle enhanced adapter tomorrow morning ET.

PR #830, opened 2016-05-06
Added :stoplist argument to add_context_index

mulvaney:
Adds a simple :stoplist argument to the add_context_index command.  The stoplist should have already been created elsewhere, this is just a way to reference the stoplist when creating the index.


gacha:
+1

yahonda:
Thanks for opening a pull request. Would you add some test cases in this file.

https://github.com/rsim/oracle-enhanced/blob/master/spec/active_record/connection_adapters/oracle_enhanced_context_index_spec.rb

such as:
* add_context_index with `STOPLIST` option creates indexes as expected
* `ActiveRecord::SchemaDumper.dump` dumps `STOPLIST` options

It validates this commit works and also should avoid future regressions.

mulvaney:
I'm not ignoring this -- I've had a lot of problems getting the current tests to run, and I haven't had much time to work on it recently.  Hopefully I'll have an update in the next couple days.

PR #835, opened 2016-05-11
Do not enable constraints in disable_referential_integrity

swaroopmurthy:
Why exactly are we trying to enable constraints in a method named disable_referential_integrity? 
We ran into some issues with transactional fixtures in a rails project, will come up with exact steps to reproduce


swaroopmurthy:
never mind

PR #882, opened 2016-07-01
Rails5

mrosel:


README.md
```diff
@@ -12,6 +12,24 @@ Oracle enhanced ActiveRecord adapter provides Oracle database access from Ruby o
 
 INSTALLATION
 ------------
+### Rails 5 - in branch rails5
+
+Rails 5 is still in pre release, so use with caution.
```
yahonda: Rails 5 itself has been released, how about "Oracle enhanced adapter Rails 5 support is ..."?

activerecord-oracle_enhanced-adapter.gemspec
```diff
@@ -82,7 +82,7 @@ This adapter is superset of original ActiveRecord Oracle adapter.
     "spec/spec_helper.rb"
   ]
   s.add_dependency(%q<rspec>, ["~> 3.3"])
-  s.add_dependency(%q<activerecord>, ["= 5.0.0.rc2"])
+  s.add_dependency(%q<activerecord>, ["~> 5"])
```
yahonda: At this moment it is unknown if we wil be able support Rails 5.0 and Rails 5.1 with the same version of Orale enhanced adapter. then how about "~> 5.0" to support only Rails 5.0.

yahonda:
Thanks for opening a pull request. I have left some comments, would you update, commit and squash if you agreed.

yahonda:
Cherry picked one commit and merged via #914. Thanks.

PR #895, opened 2016-07-14
preserve the timezone of records on database rails5

thiaguerd:
I need save and get many time zones of my records in my database, but by default I must be a choice between timezone local or utc, with this simple modification I can work with each time zone recovered of database
to use, just run this

````ruby
ActiveRecord::ConnectionAdapters::OracleEnhancedAdapter.preserve_time_zone = true
````


yahonda:
Thanks for opening a pull request, I understand you have this requirement but this is requirement looks unique to you.  Also Rails `ActiveRecord::Base.default_timezone` can only have `;utc` and `:local`, then as Oracle enhanced adapter is a adapter for Rails, I wanted to support and follow Rails configuration as much as possible. Then I prefer not to merge this pull request.

PR #896, opened 2016-07-14
preserve the timezone of records on database

thiaguerd:
I need save and get many time zones of my records in my database, but by default I must be a choice between timezone local or utc, with this simple modification I can work with each time zone recovered of database
to use, just run this

````ruby
ActiveRecord::ConnectionAdapters::OracleEnhancedAdapter.preserve_time_zone = true
````


yahonda:
Thanks for opening a pull request. Closed as #895.

Issue #851, opened 2016-06-03
 ORA-00904: invalid identifier when creating context index with dummy column name


Envek:
I'm trying to create context index with dummy column name as stated in README:

> You can also specify some dummy main column name when creating multiple column index 

But getting error `ORA-00904`

My code in migration:

```ruby
    add_context_index :address_objects,
                      [:formalname, :offname, :shortname],
                      name: 'addr_obj_search_index',
                      index_column: :all_names,
                      index_column_trigger_on: [:formalname, :offname, :shortname],
                      sync: 'ON COMMIT'
```

And error:

```
ActiveRecord::StatementInvalid: OCIError: ORA-00904: "ALL_NAMES": invalid identifier: CREATE INDEX "ADDR_OBJ_SEARCH_INDEX" ON "ADDRESS_OBJECTS" ("ALL_NAMES") INDEXTYPE IS CTXSYS.CONTEXT PARAMETERS ('DATASTORE addr_obj_search_index_dst SECTION GROUP CTXSYS.AUTO_SECTION_GROUP SYNC(ON COMMIT)')
```

There is no column `all_names` in `address_objects` table. And nowhere said that I should to create it or something.

Oracle 12
activerecord-oracle_enhanced-adapter-1.6.7
Ruby on Rails 4.2.6

yahonda:
multiple column index is tested like this and it finishes successfully.

```ruby
$  bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced_context_index_spec.rb:124
```

Here is the spec.
```ruby

    it "should create multiple column index with specified main index column" do
      @conn.add_context_index :posts, [:title, :body],
        index_column: :all_text, sync: 'ON COMMIT'
      @post = Post.create(title: "abc", body: "def")
      Post.contains(:all_text, "abc").to_a.should == [@post]
      Post.contains(:all_text, "def").to_a.should == [@post]
      @post.update_attributes!(title: "ghi")
      # index will not be updated as all_text column is not changed
      Post.contains(:all_text, "ghi").to_a.should be_empty
      @post.update_attributes!(all_text: "1")
      # index will be updated when all_text column is changed
      Post.contains(:all_text, "ghi").to_a.should == [@post]
      @conn.remove_context_index :posts, index_column: :all_text
    end
```

And posts table is defiined like this, it has `        t.string :all_text, limit: 2 # will be used for multi-column index`. 

```ruby
  def create_table_posts
    schema_define do
[...]

PR #924, opened 2016-07-21
Readme updates

Envek:
1. While upgrading to Rails 5.0 and Oracle Enhanced adapter 1.7 I've noticed deprecation warnings that ActiveRecord Attributes API should be used instead of `set_date_columns`. So, added it to examples.

 2. Document context index `index_column` option usage points from #851 

Please review for mistakes.

Thank you for Oracle Enhanced adapter. It saves us a lot of time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,6 +460,14 @@ Post.contains(:all_text, "aaa within title")
 Post.contains(:all_text, "bbb within comment_author")
 ```
 
+Please note that `index_column` must be a real column in your database and it's value will be overriden every time your `index_column_trigger_on` columns are changed. So, _do not use columns with real data as `index_column`_.
+
+Index column can be created as:
+
+```ruby
+add_column :posts, :all_text, :string, limit: 2, comment: 'Service column for context search index'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -258,14 +258,16 @@ class Employee < ActiveRecord::Base
   # specify sequence name
   self.sequence_name = "hr.hr_employee_s"
 
-  # set which DATE columns should be converted to Ruby Date
-  set_date_columns :hired_on, :birth_date_on
+  # set which DATE columns should be converted to Ruby Date using ActiveRecord Attribute API
+  attribute :hired_on, :date
+  attribute :birth_date_on, :date
```
yahonda: Starting from Oracle enhanced adapter 1.7,  Oracle `DATE` column is mapped to Rails `Date` by default.  then I am thinking about if these attribute settings should be documented here. I'm going to add this information in the comment.

yahonda:
Thanks for opening the pull request, document updates are really useful.  Thanks.

PR #981, opened 2016-08-18
[QUESTION] enable querying on CLOB/:text fields

Tuxified:
Hi,

I've created this PR as a question, as I find it easier to discuss/describe what I'd like to do.

Currently(v1.6.7) when querying a Rails `:text` field (which equals to CLOB type in Oracle), it's always matched against `empty_clob()` which only matches in case one was looking for empty fields but ignoring 'normal' cases (like `where(text_field: "long description")`.

As I'm not familiar with Oracle's data types, I was wondering if this was intentional. I'd very much like to use Rails' `:text` type, but it looses value if one cannot search for it's content. Is the change this PR is nudging to welcome (if so I'll create a proper PR with tests etc)?

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -613,7 +613,9 @@ module ActiveRecord
       def quote(value, column = nil) #:nodoc:
         if value && column
           case column.type
-          when :text, :binary
+          when :text
+            super
+          when :binary
             %Q{empty_#{ type_to_sql(column.type.to_sym).downcase rescue 'blob' }()}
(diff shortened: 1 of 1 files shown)


yahonda:
Thanks for opening a pull request. Rails `:text` is mapped to Oracle `CLOB` data type which requires `empty_clob()` operation. 

Would you update your pull request to include specs/tests which show your expected behaviour? although it would be difficult to remove this function. 

Also please consider to open a pull request to master branch, not `release-16` branch. Usually behaviour changes need to go to `master` branch first
to keep these changes affected to newer version of Oracle enhanced adapter, also older version such as 1.6.x should be stable for all users.

Thanks.

Tuxified:
Hi, 

I'll close this PR as I've created a new (couldn't change base branch) to give an example of usage which currently isn't possible.

Issue #798, opened 2016-04-04
NoMethodError: undefined method `write' for nil:NilClass for serialized column


Bjoernsen:
I am not sure, but I might have found a :bug: 

I have a table where I would like to have a column to store an Array.

So I added let's say `dents` to `cars`.
```ruby
class AddDentsToCar < ActiveRecord::Migration
  def change
    add_column :cars, :dents, :text
  end
end
```

Then, I modified the `Car` class to handle the serialization of the Array for me.
```ruby
class Cars < ActiveRecord::Base
  serialize :dents, Array
end
```

In principle, it works. I am able to play around with my `dents` Array. The issue is that I am not able to store an empty `Array` in the DB.
```ruby
car = Car.first
car.dents # => []
car.dents << 'street-lamp'
car.dents # => ['street-lamp']
car.save! # => true

car = Car.first
car.dents # => ['street-lamp']
car.dents = []
car.save! # => NoMethodError: undefined method `write' for nil:NilClass
```

For my Issue, I found a solution. But I do not know, if it breaks the adapter :smirk: 

I changed the line 828 in `oracle_enhanced_adapter.rb`, see
https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L828

from
```ruby
next if value.nil?  || (value == '')
```

to
```ruby
next if value.blank?
```

Now, it works as expected.
```ruby
car = Car.first
car.dents # => ['street-lamp']
car.dents = []
car.save! # =>true

car = Car.first
car.dents # => []
```

Environment
- ruby: `2.3.0`
- rails: `4.2.6`
- activerecord-oracle_enhanced-adapter: `1.6.7`

The log

```ruby
NoMethodError: undefined method `write' for nil:NilClass
	from /home/bjoernsen/.rvm/gems/ruby-2.3.0@dent_cars/gems/activerecord-oracle_enhanced-adapter-1.6.7/lib/active_record/connection_adapters/oracle_enhanced/oci_connection.rb:213:in `write_lob'
	from /home/bjoernsen/.rvm/gems/ruby-2.3.0@dent_cars/gems/activerecord-oracle_enhanced-adapter-1.6.7/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb:838:in `block (2 levels) in write_lobs'
[...]

PR #1005, opened 2016-10-11
undefined method for serialized column fix

Bjoernsen:


Bjoernsen:
This is the master branch fix for #798

By accident, there is another commit included. There, I added information how to use a docker Oracle XE container for the database.
Using the docker approach, some tests are failing.

yahonda:
Thanks for opening a pull request. Would you make two separate pull requests since both commit have its own purposes.

PR #1016, opened 2016-10-17
Multi insert is not supported

jlahtinen:
Dump schema migration versions with multiple SQL statement. Otherwise rails default adapter tries to add all version to single SQL statement which does not work with Oracle.

Should not affect any other functionality.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -483,6 +483,10 @@ module ActiveRecord
         true
       end
 
+      def supports_multi_insert?
+        false
+      end
+
       #:stopdoc:
       DEFAULT_NLS_PARAMETERS = {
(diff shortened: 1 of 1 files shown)


jlahtinen:
This is not enough to get rake db:test:prepare to work with `config.active_record.schema_format = :sql` but maybe one step further.

yahonda:
```sql

SQL> desc schema_migrations
 Name                                      Null?    Type
 ----------------------------------------- -------- ----------------------------
 VERSION                                   NOT NULL VARCHAR2(255 CHAR)

SQL> select * from schema_migrations;

no rows selected

SQL> INSERT INTO schema_migrations (version) VALUES ('20120321020431'), ('20120321184254');
INSERT INTO schema_migrations (version) VALUES ('20120321020431'), ('20120321184254')
                                                                 *
ERROR at line 1:
ORA-00933: SQL command not properly ended


SQL>
```

yahonda:
Thanks.

PR #1026, opened 2016-10-26
Using the gem in non-rails apps

algodave:
I'm using this gem in a sinatra app with ActiveRecord out of the Rails box.
I think it's worth to keep a `if defined?(Rails)` condition before `OracleEnhancedRailtie` definition for those like me who want to use this great gem in non-rails apps.


yahonda:
Thanks. it is kind of reverting #769.

PR #1032, opened 2016-11-03
Support connection strings in DATABASE_URL

jbotelho2-bb:
We are trying to move our Rails apps to stick to using environment variables for configuration, and noticed that we could not put a connection string in the url.

To solve this, I added a simple addition to support putting a connection-string in the URL (as the path), when the host is `connection-string`. This change seemed fairly simple and unlikely to break existing configurations, but I'm open to other ways to pick up on the connection string from there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,12 @@ environment variable, note that the adapter name uses a dash instead of an under
 DATABASE_URL=oracle-enhanced://localhost/XE
 ```
 
+You can also specify a connection string via the `DATABASE_URL`, as long as it doesn't have any whitespace:
+
+```bash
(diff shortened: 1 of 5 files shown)


yahonda:
Thanks for opening a pull request. would you add specs to validate the `DATABASE_URL=oracle-enhanced://user:secret@connection-string/(DESCRIPTION=(ADDRESS_LIST=(ADDRESS=(PROTOCOL=tcp)(HOST=localhost)(PORT=1521)))(CONNECT_DATA=(SERVICE_NAME=xe)))` works for both OCI and JDBC connections.

Thanks.

jbotelho2-bb:
I added JRuby/JDBC support and a spec to cover both platforms (OCI and JDBC).

yahonda:
Thanks.

jbotelho2-bb:
@yahonda thanks for merging this.

Would it be possible to get this change into the 1.6 release as well? Should I make a pull request for that?

yahonda:
#1036 has been merged to release16 branch.

PR #1077, opened 2016-12-06
Fallback :bigint to :integer for primary keys

yahonda:
Since Oracle enhanced adapter already handles Rails primary keys as NUMBER(38) Oracle data type.
Here 38 means precision, not bytes to store.

Also Rails :bigint data type is handled as NUMBER(19) Oracle data type which has less precision
than Oracle enhanced adapter :integer. Then this pull request falls back :bigint for primary keys
to Rails :integer.

Refer rails/rails#26266

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/column_dumper.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/column_dumper.rb
@@ -53,6 +53,10 @@ module ActiveRecord #:nodoc:
 
         private
 
+        def default_primary_key?(column)
+          schema_type(column) == :integer
+        end
+
         def schema_virtual_as(column)
           column.virtual_column_data_default if column.virtual?
         end
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_definitions.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_definitions.rb
@@ -1,6 +1,18 @@
 module ActiveRecord
   module ConnectionAdapters
     module OracleEnhanced
+      class ReferenceDefinition < ActiveRecord::ConnectionAdapters::ReferenceDefinition # :nodoc:
+        def initialize(
+          name,
+          polymorphic: false,
+          index: true,
+          foreign_key: false,
+          type: :integer,
+          **options)
+          super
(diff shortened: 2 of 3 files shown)


yahonda:
Also this pull request addresses this failure:

```ruby
$ ARCONN=oracle bundle exec ruby -w -Itest test/cases/schema_dumper_test.rb -n test_schema_dump_uses_force_cascade_on_create_table
... snip ...
Run options: -n test_schema_dump_uses_force_cascade_on_create_table --seed 34771

# Running:

/home/yahonda/.rbenv/versions/2.4.0-dev/lib/ruby/gems/2.4.0/gems/ruby-oci8-2.2.2/lib/oci8/oci8.rb:625: warning: constant ::Fixnum is deprecated
F

Finished in 38.597733s, 0.0259 runs/s, 0.0518 assertions/s.

  1) Failure:
SchemaDumperTest#test_schema_dump_uses_force_cascade_on_create_table [test/cases/schema_dumper_test.rb:61]:
Expected /create_table "authors", force: :cascade/ to match "# This file is auto-generated from the current state of the database. Instead\n# of editing this file, please use the migrations feature of Active Record to\n# incrementally modify your database, and then regenerate this schema definition.\n#\n# Note that this schema.rb definition is the authoritative source for your\n# database schema. If you need to create the application database on another\n# system, you should be using db:schema:load, not running all the migrations\n# from scratch. The latter is a flawed and unsustainable approach (the more migrations\n# you'll amass, the slower it'll run and the greater likelihood for issues).\n#\n# It's strongly recommended that you check this file into your version control system.\n\nActiveRecord::Schema.define(version: 0) do\n\n  create_table \"authors\", id: :integer, precision: 38, force: :cascade do |t|\n    t.string  \"name\",                                              null: false\n    t.integer \"author_address_id\",       limit: 19, precision: 19\n    t.integer \"author_address_extra_id\",            precision: 38\n    t.string  \"organization_id\"\n    t.string  \"owned_essay_id\"\n  end\n\n  add_foreign_key \"authors\", \"author_addresses\"\nend\n".

1 runs, 2 assertions, 1 failures, 0 errors, 0 skips
$
```

PR #1168, opened 2017-02-10
Add Virtual column support without type

yahonda:
This pull request follows up #1167 to restore Virtual column support without type.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_creation.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_creation.rb
@@ -5,7 +5,7 @@ module ActiveRecord
         private
 
           def visit_ColumnDefinition(o)
-            if [:blob, :clob].include?(sql_type = type_to_sql(o.type.to_sym,  o.options).downcase.to_sym)
+            if [:blob, :clob].include?(sql_type = type_to_sql(o.type,  o.options).downcase.to_sym)
               if (tablespace = default_tablespace_for(sql_type))
                 @lob_tablespaces ||= {}
                 @lob_tablespaces[o.name] = tablespace


lib/active_record/connection_adapters/oracle_enhanced/schema_creation.rb
```diff
@@ -5,10 +5,12 @@ class SchemaCreation < AbstractAdapter::SchemaCreation
         private
 
           def visit_ColumnDefinition(o)
-            if [:blob, :clob].include?(sql_type = type_to_sql(o.type.to_sym,  o.options).downcase.to_sym)
-              if (tablespace = default_tablespace_for(sql_type))
-                @lob_tablespaces ||= {}
-                @lob_tablespaces[o.name] = tablespace
+            unless o.type.nil?
+              if [:blob, :clob].include?(sql_type = type_to_sql(o.type.to_sym,  o.options).downcase.to_sym)
```
kamipo: I think that `to_sym` to `o.type` is unnecessary.
https://github.com/rails/rails/blob/55d66e214825b89442383f90adf575cf8c96568c/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L1056

kamipo: Originally I intended that `virtual` without `type` is to be `nil.to_s`.
https://github.com/rails/rails/blob/55d66e214825b89442383f90adf575cf8c96568c/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L1085

yahonda: Thanks for the advice. I got it.  Updated this pull request.

PR #1181, opened 2017-02-14
`release_savepoint` is not supported by Oracle

yahonda:
Better to raise `NotImplementedError` rather than do nothing


koic:
@yahonda Thanks for the maintenance. Share the problem I noticed.

With this change, More than 1,000 erros in AR of rails/rails tests by Oracle.
I think that it is only the impact on the test (not production).

This is just a quick note, just in case.

yahonda:
It's my bad. Thanks for letting me know. #1182 has been opened.

koic:
Thank you for the revert patch.

PR #1208, opened 2017-02-26
Push `valid_type?` up to abstract adapter

koic:
Follow up to https://github.com/rails/rails/pull/28176.

It becomes the same implementation as the super class, it was removed.


yahonda:
Also, let me revert this commit temporary to release Oracle enhanced adapter 1.8.0.beta1. Once it is released I'll revert this revert. Thanks always catching up the latest version of Rails.

koic:
Also, thank you for this.

PR #1209, opened 2017-02-27
Deprecate `supports_migrations?` on connection adapters

koic:
Follow up to https://github.com/rails/rails/pull/28172.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -263,10 +263,6 @@ module ActiveRecord
         end
       end
 
-      def supports_migrations? #:nodoc:
-        true
-      end
-
       def supports_savepoints? #:nodoc:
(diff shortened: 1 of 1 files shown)


yahonda:
Good catch as always. This change has been made after Rails 5.1.0.beta1 was released. 

We have not released a beta version of Oracle enhanced adapter 1.8, then I am going to revert this change temporary to release Oracle enhanced adapter 1.8.0.beta1 and restore this commit.

As long as Rails 5.1 beta version is released from master branch, we will release Oracle enhanced adapter 1.8.0.beta from our master branch. Once Rails branch `5-1-stable` (or equivalent branch) made, we will create a new branch `release18`.

koic:
> This change has been made after Rails 5.1.0.beta1

oops! I overlooked that.
I understood about this maintenance policy. And, thanks for the Oracle enhanced adapter 1.8.0.beta1 release.

I'm always thankful for the help.

PR #1220, opened 2017-03-04
Suppress `WARNING: Using `expect { }.not_to raise_error(...)` risks false positives`

koic:
This PR suppresses the following warning.

```sh
% bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced_dirty_spec.rb:128

(snip)

WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass, message)` risks false positives, since literally any other error would cause 
the expectation to pass, including those raised by Ruby (e.g. NoMethodError, NameError and ArgumentError), meaning the code you are intending to t
est may not even get reached. Instead consider using `expect {}.not_to raise_error` or `expect { }.to raise_error(DifferentSpecificErrorClas
s)`. This message can be suppressed by setting: `RSpec::Expectations.configuration.on_potential_false_positives = :nothing`. Called from /home/
vagrant/src/oracle-enhanced/spec/active_record/connection_adapters/oracle_enhanced_dirty_spec.rb:128:in `block (2 levels) in <top (required)>'.
.

Finished in 0.15775 seconds (files took 0.51207 seconds to load)
1 example, 0 failures
```

### DIFF
--- a/spec/active_record/connection_adapters/oracle_enhanced_dirty_spec.rb
+++ b/spec/active_record/connection_adapters/oracle_enhanced_dirty_spec.rb
@@ -125,7 +125,7 @@ if ActiveRecord::Base.method_defined?(:changed?)
       class << oci_conn
          def write_lob(lob, value, is_binary = false); raise "don't do this'"; end
       end
-      expect { @employee.save! }.not_to raise_exception(RuntimeError, "don't do this'")
+      expect { @employee.save! }.not_to raise_error
       class << oci_conn
         remove_method :write_lob
       end


yahonda:
Thanks for opening a pull request. I had attempted to address this by raising another specified error.
as suggested.

```ruby
consider using ... snip ...`expect { }.to raise_error(DifferentSpecificErrorClass)`. 
```

I would prefer not to remove error message here. Do you have some idea how to suppress this warning by changing as above?

koic:
Thanks for your review.

```sh
Instead consider using `expect {}.not_to raise_error` or `expect { }.to raise_error(DifferentSpecificErrorClass)`.
```

The reason why I chose `expect {}.not_to raise_error` of the above is because  I thought that I expected no error at all in this test case. Therefore, I removed the specific error class and message.

> Do you have some idea how to suppress this warning by changing as above?

I chose the `expect { }.to raise_error(DifferentSpecificErrorClass)` if it expects any exceptions to occur. But I don't have this idea now.

yahonda:
Thanks for the update.

koic:
Thanks too.

PR #1240, opened 2017-03-17
Add JSON attribute support

yahonda:
This pull request adds JSON attribute support.

- Oracle Database 12c Release 1 or higher version is required

- ActiveRecord migration for json type is not supported
since there is no JSON data type in Oracle database

1. Define :string or :text in migration for columns for JSON
```ruby
  create_table :test_posts, force: true do |t|
    t.string  :title
    t.text    :article
  end
```

2. Set :json attributes

```ruby
  class TestPost < ActiveRecord::Base
    attribute :title, :json
    attribute :article, :json
  end
```

3. Add `is json` database constraints by running sql statements
```sql
  alter table test_posts add constraint test_posts_title_is_json check (title is json)
  alter table test_posts add constraint test_posts_article_is_json check (article is json)
```


deepj:
Is it possible to mention this in the documentation?

Issue #1244, opened 2017-03-20
BINARY_FLOAT return nil (Rails 2.4 with oracle-enhanced 1.6.8)


shozayen:
I am using Rails 4.2.7 with oracle adapter 1.6.8. I can save the data into oracle DB as float without any issue but when I retrieve the data, the data in Float columns returned as **nil**. 

**Image:**  
[https://i.stack.imgur.com/yA2GI.png](https://i.stack.imgur.com/yA2GI.png)

yahonda:
Thanks for your report. Would you create a executable test case based on this template below.

```ruby
begin
  require 'bundler/inline'
rescue LoadError => e
  $stderr.puts 'Bundler version 1.10 or later is required. Please update your Bundler'
  raise e
end

gemfile(true) do
  source 'https://rubygems.org'
  gem 'rails', '~> 4.2.0'
  gem 'activerecord-oracle_enhanced-adapter', '~> 1.6.0'
  gem 'ruby-oci8'
  gem 'minitest'
end

require 'active_record'
require 'minitest/autorun'
require 'logger'
require 'active_record/connection_adapters/oracle_enhanced_adapter'

# Ensure backward compatibility with Minitest 4
Minitest::Test = MiniTest::Unit::TestCase unless defined?(Minitest::Test)

# Set Oracle enhanced adapter specific connection parameters
DATABASE_NAME = ENV['DATABASE_NAME'] || 'orcl'
DATABASE_HOST = ENV['DATABASE_HOST']
DATABASE_PORT = ENV['DATABASE_PORT']
DATABASE_USER = ENV['DATABASE_USER'] || 'oracle_enhanced'
DATABASE_PASSWORD = ENV['DATABASE_PASSWORD'] || 'oracle_enhanced'
DATABASE_SYS_PASSWORD = ENV['DATABASE_SYS_PASSWORD'] || 'admin'

CONNECTION_PARAMS = {
  :adapter => "oracle_enhanced",
  :database => DATABASE_NAME,
  :host => DATABASE_HOST,
  :port => DATABASE_PORT,
  :username => DATABASE_USER,
  :password => DATABASE_PASSWORD
}

ActiveRecord::Base.establish_connection(CONNECTION_PARAMS)

ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :posts, :force => true do |t|
  end

  create_table :comments, :force => true do |t|
    t.integer :post_id
  end
end

[...]

PR #1254, opened 2017-03-26
Fix the BINARY_FLOAT column type returns nil in JRuby

koic:
Refer #1244.

This PR will return an approximate value instead of nil when using BINARY_FLOAT column type in OracleEnhancedJDBCConnection. That is the same behavior as OracleEnhancedOCIConnection.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
@@ -506,6 +506,8 @@ module ActiveRecord
           else
             BigDecimal.new(d.stringValue)
           end
+        when :BINARY_FLOAT
+          rset.getFloat(i)
         when :VARCHAR2, :CHAR, :LONG, :NVARCHAR2, :NCHAR
           rset.getString(i)
         when :DATE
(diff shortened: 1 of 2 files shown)


yahonda:
Really great job for JRuby support.

shozayen:
@koic Thanks a lot

koic:
@shozayen You're welcome!

yahonda:
@shozayen 

Would you confirm if `float` is a appropriate data type for your requirements.

As #1254 test case describes `4.4` becomes `4.400000095367432` once saved to Rails `float` type associated with Oracle `BINARY FLOAT` sql type.

Let me quote Oracle documents for the reason:

[Floating-Point Numbers
](https://docs.oracle.com/database/121/SQLRF/sql_elements001.htm#SQLRF0021
)
>

> Floating-Point Numbers
> Floating-point numbers can have a decimal point anywhere from the first to the last digit or can have no decimal point at all. An exponent may optionally be used following the number to increase the range, for example, 1.777 e-20. A scale value is not applicable to floating-point numbers, because the number of digits that can appear after the decimal point is not restricted.
> 
> Binary floating-point numbers differ from NUMBER in the way the values are stored internally by Oracle Database. Values are stored using decimal precision for NUMBER. All literals that are within the range and precision supported by NUMBER are stored exactly as NUMBER. Literals are stored exactly because literals are expressed using decimal precision (the digits 0 through 9). Binary floating-point numbers are stored using binary precision (the digits 0 and 1). Such a storage scheme cannot represent all values using decimal precision exactly. Frequently, the error that occurs when converting a value from decimal to binary precision is undone when the value is converted back from binary to decimal precision. The literal 0.1 is such an example.

Let me confirm if your application requires `float` data type, not `decimal` data type.

Thanks.

shozayen:
@yahonda 
My Application needs float as I am storing longitude and latitude in my Oracle DB. If I use binary_float it is doing a slight change to the value, I tried the Float column in Oracle with default precision (127) and it seems good as it is not changing the value.

yahonda:
@shozayen Thanks for the update.

PR #1255, opened 2017-03-26
Fix the BINARY_FLOAT column type returns nil in JRuby

yahonda:
This pull request back ports #1254 to release18 branch.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
@@ -506,6 +506,8 @@ module ActiveRecord
           else
             BigDecimal.new(d.stringValue)
           end
+        when :BINARY_FLOAT
+          rset.getFloat(i)
         when :VARCHAR2, :CHAR, :LONG, :NVARCHAR2, :NCHAR
           rset.getString(i)
         when :DATE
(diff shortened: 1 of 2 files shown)


koic:
@yahonda Thanks for your backporting.
I will also going to backport this to the `release17` branch and the `release16` branch. How about this one?

yahonda:
I have not decided to backport to release17 and release16 yet. The main reasons are:
* I'd like to reduce backports especially for release16 branch supporting Rails 4.2. 
* I'd like to confirm if `float` is an appropriate data type for the original issue. Then updated a comment in #1244  

Will update when I've decided to go or no go.

koic:
I got it. I will pause this backporting. Thanks for your reply.

yahonda:
@koic Would you open pull requests to backport this change into release17 and release16 branch, also release 1.7.11 and 1.6.9 if you do not mind.

koic:
@yahonda Thanks for your decision. I'll be happy to do it.

lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
```diff
@@ -506,6 +506,8 @@ def get_ruby_value_from_result_set(rset, i, type_name, get_lob_value = true)
           else
             BigDecimal.new(d.stringValue)
           end
+        when :BINARY_FLOAT
+          rset.getFloat(i)
```
tillsc, line 510: This introduces a new bug: `NULL` values will be always casted to `0.0` since [getFloat](https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#getFloat(int)) never returns `nil`.

A quick fix could look like:

    rset.getString(i) && rset.getFloat(i)

(since [getString](https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#getString(int)) returns `nil` for `NULL` values)

tillsc: Created issue #2040 regarding this problem

PR #1260, opened 2017-03-29
Bump Ruby version to 2.2.7

yahonda:
https://www.ruby-lang.org/en/news/2017/03/28/ruby-2-2-7-released/


yahonda:
Not available yet, Refer rvm/rvm#3970

PR #1270, opened 2017-03-30
Use Ruby 2.3.4

yahonda:
https://www.ruby-lang.org/en/news/2017/03/30/ruby-2-3-4-released/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ install:
 language: ruby
 rvm:
   - 2.2.7
-  - 2.3.3
+  - 2.3.4
   - 2.4.1
   - ruby-head
   - jruby-9.1.8.0


yahonda:
Refer rvm/rvm#3973.  Would like to know if there is a way to use the latest `TEENY` version of ruby.

yahonda:
@koic Do you have some idea or request at RVM or Travis to use the latest `TEENY` version of Ruby? 
I'd like to test with the latest version of Ruby without bumping teeny version every time.

koic:
@yahonda As far as I know, Ruby 2.2 and Ruby 2.3 can omit the TEENY version.

```Diff
- - 2.2.7
- - 2.3.4
- - 2.4.1
+ - 2.2
+ - 2.3
+ - 2.4.1
```

However, I do not know whether it always works with the latest TEENY version.

In addition, if the TEENY version is explicitly written, the tested version may become obvious to users (or developers) .

yahonda:
Thanks for the information. Also there used to be my desired feature but it was rolled back https://github.com/travis-ci/travis-ci/issues/5581 . Let's specify and update the latest teeny version.

PR #1295, opened 2017-04-18
Follow the new interface of AR::ConnectionAdapters::IndexDefinition#initialize

koic:
Refer: rails/rails#28773.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_definitions.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_definitions.rb
@@ -42,7 +42,7 @@ module ActiveRecord
           @parameters = parameters
           @statement_parameters = statement_parameters
           @tablespace = tablespace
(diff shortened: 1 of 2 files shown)


yahonda:
Good catch as always. rails/rails#28773 has two commits and this pull request addresses the first one https://github.com/rails/rails/pull/28773/commits/1a92ae8318c3f5720907dec671d9ebb9221c0817 . 

for the second one it implements descending index for MySQL 8.0.1 and later. As far as I remember, Oracle database itself already supports descending index as functional index. but Oracle enhanced adapter does not consider it since it does not have `supports_index_sort_order?` method.

koic:
Thank you for your explanation. I understand about the positioning of the `supports_index_sort_order?` method in Oracle enhanced adapter.

yahonda:
Off course pull requests are welcomed but I do not use descending order indexes then I am not just feeling like to implement this feature myself. I assume MySQL relies on indexes for `order by` clause than Oracle.

PR #1315, opened 2017-05-01
Install probot

yahonda:
This ticket requests to enable https://github.com/probot/stale to manage old issues.
Changed to close only stale issues, not old requests.


yahonda:
@rsim I have added you as reviewers. Once you agreed with this pull request please enable probot.

Thanks.

PR #1317, opened 2017-05-01
[ci skip] Create an issue template

yahonda:
Referred the one from Rails repository https://github.com/rails/rails/blob/master/.github/issue_template.md and add necessary sections for Oracle enhanced adapter.

This time [ci skip] subject added to the commit message, not the subject of pull request. Will see how it goes.


yahonda:
[ci skip] works now.

Issue #1291, opened 2017-04-07
two database tasks spec failure when tested with JRuby at Travis


yahonda:
Not always but sometimes these specs get failed when tested with JRuby 9.1.8.0

https://travis-ci.org/rsim/oracle-enhanced/jobs/219757312

```ruby
rspec ./spec/active_record/connection_adapters/oracle_enhanced_database_tasks_spec.rb:80 # Oracle Enhanced adapter database tasks with test table structure structure_dump dumps the database structure to a file without the schema information
rspec ./spec/active_record/connection_adapters/oracle_enhanced_database_tasks_spec.rb:93 # Oracle Enhanced adapter database tasks with test table structure structure_load loads the database structure from a file
```

I do not see these failures locally.

```
Oracle Database 12c Enterprise Edition Release 12.2.0.1.0 - 64bit Production
$ java -version
openjdk version "1.8.0_121"
OpenJDK Runtime Environment (build 1.8.0_121-b14)
OpenJDK 64-Bit Server VM (build 25.121-b14, mixed mode)
$ ruby -v
jruby 9.1.8.0 (2.3.1) 2017-03-06 90fc7ab OpenJDK 64-Bit Server VM 25.121-b14 on 1.8.0_121-b14 [linux-x86_64]
$ echo $JRUBY_OPTS
--debug --dev -J-Xmx1024M
```

yahonda:
Since reverting to Ubuntu 12.04, these failures do not reproduce, possibly due to slow tests with Ubuntu 14.04. 

Closing it for now and let's revisit if reproduces again after we migrate to 14.04 again.

PR #1342, opened 2017-05-17
CI against JRuby 9.1.9.0

koic:
JRuby 9.1.9.0 has been released.

http://jruby.org/2017/05/16/jruby-9-1-9-0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ rvm:
   - 2.3.4
   - 2.4.1
   - ruby-head
-  - jruby-9.1.8.0
+  - jruby-9.1.9.0
   - jruby-head
 
 matrix:


yahonda:
CI for jruby-9.1.9.0 and ruby-head is failing due to #1291. 

There is a similar error generated at [simplecov](https://github.com/colszowka/simplecov) .
https://travis-ci.org/colszowka/simplecov/jobs/232794265

yahonda:
Opened https://github.com/colszowka/simplecov/pull/582 to run CI against Ruby 9.1.0.0 to see if it reproduces with simplecov itself.

koic:
Thanks for your advice.
It seems like a bug in JRuby 9.1.9.0. I will open an issue of JRuby later.

koic:
I think that this build error will fix with the following PR.

colszowka/simplecov-html#53

I'd like to see how it goes for a while. How about this?

yahonda:
Thanks for the investigation.
I'd like to see how https://github.com/jruby/jruby/issues/4610 goes. Since this  behavior change looks a regression for me which would affect many other users.

koic:
Sure. I got it.

koic:
It was a regression of JRuby 9.1.9.0. This regression issue was included in the milestone of JRuby 9.2.0.0.

https://github.com/jruby/jruby/milestone/53

If they have another regressions it is even better to test with them (That's fail fast) . So I think that it is good to merge this PR. Tell me what you think.

yahonda:
Really great job. Thank you.

koic:
Thanks too.

PR #1354, opened 2017-06-13
Add sequence with settings to structure dump

jlahtinen:
Add min and max values, order and cycle flags.

These are sometimes needed when sequence is not solely used for id generation for example.

order flag is quite useful also sometimes for id generation.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/structure_dump.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/structure_dump.rb
@@ -5,8 +5,8 @@ module ActiveRecord #:nodoc:
       STATEMENT_TOKEN = "\n\n/\n\n"
 
       def structure_dump #:nodoc:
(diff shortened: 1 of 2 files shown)


yahonda:
Thanks for opening a pull request. Do you think you need `CACHE_SIZE` to set the cache size of the sequence?

jlahtinen:
@yahonda I think we do not need it. It would be no harm anyway.

yahonda:
Thanks for the answer. This pull request has been merged.

jlahtinen:
@yahonda  thank you for keeping this project alive. When you make a new release? Is there any policy for that to read?

yahonda:
This is a new feature then it will be available for 1.9.0 to support Rails 5.2. Not for Rails 5.1 or lower version.
Since it changes the output of the schema dump file, this kind of change is not suitable for the maintenance release.

PR #1364, opened 2017-06-20
Mention new statement_limit default

jlahtinen:
I think this makes it easier to spot changed statement_limit value when upgrading


yahonda:
Thank you.

PR #1373, opened 2017-07-11
Enable Layout/EmptyLineAfterMagicComment cop

yahonda:
Follow Rails implementation rails/rails#29748 No offenses found so far.


koic:
Thanks for the catch.

PR #1374, opened 2017-07-14
Enable frozen string literal

koic:
## Summary

It seems that introduction of frozen string literal to Rails has started, so this PR applies it to Oracle enhanced adapter.

https://github.com/rails/rails/blob/472f39f781e5ebf5dcb19bb432572ba68dfcceca/.rubocop.yml#L83-L88

I also confirmed that the "`can not modify frozen String`" error does not occur in rails/activerecord tests.

## Other Information

The "`can not modify frozen String`" error is fixed by `String#dup`. I think that `String#dup` is preferable to `String.new` for the following reasons.

https://github.com/rails/rails/pull/29506#issuecomment-309672824

There are the following PR recently merged into RuboCop.

https://github.com/bbatsov/rubocop/pull/4586

It seems that `String#+@` is superior to `String#dup`, but this PR does not using it because support target can not be limited to Ruby 2.3 or higher.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -74,6 +74,10 @@ Layout/SpaceBeforeFirstArg:
 Style/MethodDefParentheses:
   Enabled: true
 
+Style/FrozenStringLiteralComment:
+  Enabled: true
+  EnforcedStyle: always
+
 # Use `foo {}` not `foo{}`.
 Layout/SpaceBeforeBlockBraces:
   Enabled: true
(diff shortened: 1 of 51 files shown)


yahonda:
Refer rails/rails#29732

yahonda:
Thanks for the work. Let's see rails/rails#29732 and decide to merge it or not.

koic:
I got it. Thanks!

PR #1379, opened 2017-07-22
Update documentation that supported Oracle version [ci skip]

koic:
This PR update documentation that Oracle enhanced adapter also work with Oracle 12c.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# oracle_enhanced_adapter.rb -- ActiveRecord adapter for Oracle 8i, 9i, 10g, 11g
+# oracle_enhanced_adapter.rb -- ActiveRecord adapter for Oracle 10g, 11g and 12c
 #
 # Authors or original oracle_adapter: Graham Jenkins, Michael Schoen
 #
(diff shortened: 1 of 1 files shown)


lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
```diff
@@ -103,7 +103,7 @@ module ConnectionAdapters #:nodoc:
     # Ruby 1.8/1.9 ruby-oci8 gem (which provides interface to Oracle OCI client)
     # or with JRuby and Oracle JDBC driver.
     #
-    # It should work with Oracle 9i, 10g and 11g databases.
+    # It should work with Oracle 9i, 10g, 11g and 12c databases.
     # Limited set of functionality should work on Oracle 8i as well but several features
     # rely on newer functionality in Oracle database.
```
koic: @yahonda If Oracle enhanced adapter already does not support Oracle 8i and 9i, I think I will remove it from here. What do you think?

yahonda: Recent changes do not consider older version of Oracle 8i or 9i, such as virtual columns and TIMESTAMP data type . Then let's remove documentation mentioning 9i or older versions of Oracle.

koic: Thanks for replying. I updated and rebased this PR.

yahonda:
Thanks.

PR #1380, opened 2017-07-24
Add synonyms in data_sources

adorechic:
I'm wondering why `data_sources` does not returns synonyms.
Current `data_sources` and `data_source_exist?` causes different behavior.

```ruby
conn.data_source_exist?('my_synonym') # => true
conn.data_sources.include?('my_synonym') # => false
```

I think synonyms are available as data source for AR models, so it is reasonable that `data_sources` includes synonyms.
Are there any reason to not treat synonyms as `data_sources` ?

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -527,7 +527,7 @@ module ActiveRecord
       end
 
       def data_sources
-        super
+        super | synonyms.map(&:name)
       end
 
       def table_exists?(table_name)
--- a/spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
+++ b/spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
@@ -729,4 +729,29 @@ describe "OracleEnhancedAdapter" do
       expect(TestLog.count).to eq 2
     end
   end
+
+  describe "synonym_names" do
+    before(:all) do
+      schema_define do
+        create_table :test_comments, force: true do |t|
+          t.string :comment
+        end
+        add_synonym :synonym_comments, :test_comments
+      end
+    end
+
+    after(:all) do
+      schema_define do
+        drop_table :test_comments
+        remove_synonym :synonym_comments
+      end
+      ActiveRecord::Base.clear_cache! if ActiveRecord::Base.respond_to?(:"clear_cache!")
+    end
+
+    it "includes synonyms in data_source" do
+      conn = ActiveRecord::Base.connection
+      expect(conn).to be_data_source_exist("synonym_comments")
+      expect(conn.data_sources).to include("synonym_comments")
+    end
+  end
 end


yahonda:
Thanks for opening a pull request.  Would you answer these questions so that I can understand some more backgrounds for your pull request?


* I assume your implementation expects synonyms are pointing tables or views
What if synonyms are pointing non table/view objects such as PL/SQL stored procedures? (This is my biggest concern not to add synonyms for data_source)

* Does `self.table_name=` satisfy your requirments?

http://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-table_name-3D

```ruby
class Mouse < ActiveRecord::Base
  self.table_name = "mice"
end
```

adorechic:
> your implementation expects synonyms are pointing tables or views

yes,

> What if synonyms are pointing non table/view objects such as PL/SQL stored procedures?

Ah, now I see what is you're worrying about. I didn't care such case.

BTW, current `data_source_exist?` returns true for synonyms, and `SchemaDumper#tables` also dumps any synonyms. These behaviors do not cause any problems...? (This topic may be out of this issue.)

> Does self.table_name= satisfy your requirments?

No. 
My background is a little complicated.

I'm finding a way to manage my oracle schema without AR Migration. 
Now I try to use [ridgepole](https://github.com/winebarrel/ridgepole). Ridgepole works fine for most of my tables, but it does not support synonyms. I don't need to manage these synonyms in my situation, so I want to ignore these synonyms. (ridgepole supports ignore_tables option which is passed to `AR::SchemaDumper.ignore_tables`). But ridgepole [applies ignore_tables which in data_sources](https://github.com/winebarrel/ridgepole/blob/9b3362881f8438cccafef2549e8061033f14b66f/lib/ridgepole/dumper.rb#L18-L24), so ridgepole cannot ignore synonyms via oracle_enhanced.

Even if I add synonym support to ridgepole, it still cannot ignore synonyms. I have to remove ignore_tables validation with datasources from ridgepole.
But validating ignore_tables with data_sources looks not bad idea for me, so I'm wondering why oracle_enhanced does not include synonyms in datasources, and opened this pull request.

Now I see this change may have an impact on non table/view object synonyms, so I don't propose strongly. If you feel too large impact to merge this change, please close this pull request.
Thanks,

lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
```diff
@@ -527,7 +527,7 @@ def tables #:nodoc:
       end
 
       def data_sources
-        super
+        tables | views | synonym_names
```
yahonda: how about `super | synonym_names`?  Oracle enhanced adapter should not be affected in case of Abstract adapter `data_sources` implementation changes.

lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
```diff
@@ -569,6 +569,10 @@ def materialized_views #:nodoc:
         select_values("SELECT LOWER(mview_name) FROM all_mviews WHERE owner = SYS_CONTEXT('userenv', 'current_schema')")
       end
 
+      def synonym_names #:nodoc:
+        select_values("SELECT LOWER(synonym_name) FROM all_synonyms where owner = SYS_CONTEXT('userenv', 'current_schema')")
```
yahonda: I prefer this method name `synonyms` and found changing this method name to `synonyms` causes this error at Schema dumper.

```
NoMethodError: undefined method `name' for "test_synonym":String
```
I'd like to address this error and use `synonyms` method name here to have better consistency with `tables` and `views`.

adorechic: Current `OracleEnhancedAdapter` has already `synonyms` method which implemented in [`SchemeStatementsExt#synonyms`](https://github.com/rsim/oracle-enhanced/blob/abc596e7f8f78fa25cabc05b167cd8f0241e14d5/lib/active_record/connection_adapters/oracle_enhanced/schema_statements_ext.rb#L53-L58), so that `synonyms` method name conflicts with this, and this method returns `SynonymDefinition` objects. This is why I used `synonym_names` for this new method.

Instead of new method, how about this?
```ruby
def data_sources
  super | synonyms.map(&:name)
end
```

yahonda: Sounds good.  #1387 moved `synonyms` method from `SchemaStatementsExt` module to https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L573-L579 Since I overlooked `SchemaStatementsExt` module.

Would you remove `synonym_names` method here and change `data_sources` method as you suggested.

yahonda:
Once all changes made, please squash your commits into one.

spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
```diff
@@ -729,4 +729,29 @@ class TestLog < ActiveRecord::Base
...
+      schema_define do
+        drop_table :test_comments
+        remove_synonym :synonym_comments
+      end
+      ActiveRecord::Base.clear_cache! if ActiveRecord::Base.respond_to?(:"clear_cache!")
+    end
+
+    it "includes synonyms in data_source" do
+      conn = ActiveRecord::Base.connection
+      expect(conn).to be_data_source_exist("synonym_comments")
```
yahonda: Does this line want to test `data_source_exists?` method then it should be `be_data_source_exists` ? I mean the trailing `s` is missing?

yahonda: Looks like both `be_data_source_exist` and `be_data_source_exists` are working.  May be this is a common knowledge between RSpec users, but I just do not know.

adorechic: Yes. Predicate be matcher supports present tense predicate, so both expectations work.
https://github.com/rspec/rspec-expectations/blob/8c1e0e419125fe2cb1ddc8932c163dbc70e70668/lib/rspec/matchers/built_in/be.rb#L244-L250

adorechic:
I changed codes and squashed commits.
0f74a8ba544b887ba44979b3ecfc7c8b3f684ff2

yahonda:
Thank you.

PR #1381, opened 2017-07-24
Make `type_map` to private because it is only used in the connection adapter

koic:
Follow up of rails/rails#29869.


koic:
This PR fixes the following build error.
https://travis-ci.org/rsim/oracle-enhanced/builds/256638703

PR #1385, opened 2017-07-25
Remove add_runtime_dependency with arel

yahonda:
Oracle enhanced adapter depends on ActiveRecord.
ActiveRecord depends on Arel which is described in activerecord/activerecord.gemspec

Related with https://github.com/rails/rails/commit/089ca520e32e8cd9ebe1075f491d7856683e6cde


yahonda:
@koic I would like you to review this pull request since you have better background knowledge for this kind of changes.

koic:
Thank you for good solution.

yahonda:

PR #1396, opened 2017-07-31
`.freeze` is not necessary anymore since `# frozen_string_literal: true`

yahonda:
Refer #1374


yahonda:
Thanks for the review.  You are right.  `# frozen_string_literal: true`  just works for Ruby 2.3 or higher.
https://www.ruby-lang.org/en/news/2015/12/25/ruby-2-3-0-released/

As long as Rails 5.2 supports Ruby 2.2, Oracle enhanced adapter 1.9 will need to support it too.
Also, I do not have good benchmark tool to compare performance (memory usage?). Let's keep the current `String#freeze`.

koic:
I see. I think that these affects memory and GC, but I don't know the actual impact to a Ruby 2.2 application, so I think it is a good decision. Thank you!

Issue #1397, opened 2017-07-31
CI on Travis getting slower on Trusty


yahonda:
### Steps to reproduce
Run Oracle enhanced adapter specs using Travis CI. One is on Ubuntu Trusty another one is on Ubuntu Precise.

### Expected behavior
Unit tests using Trusty should be as fast as Precise.

### Actual behavior
Unit tests using Precise are slower than Trusty about 2 to 2.5 times.

#### Trusty, will be new default

https://travis-ci.org/rsim/oracle-enhanced/builds/259379061

- CRuby about 8 min
- JRuby about 26 min

#### Precise the old default
https://travis-ci.org/rsim/oracle-enhanced/builds/258648463

- CRuby about 4 min
- JRuby about 10 min

### System configuration
**Rails version**:master branch

**Oracle enhanced adapter version**:master branch

**Ruby version**: CRuby 2.2.1 / 2.3.4 / 2.4.1 / ruby-trunk and JRuby 9.1.12 / jruby-head

**Oracle Database version**: Oracle Database 11g XE

yahonda:
Closed by #1413

PR #1398, opened 2017-07-31
Set `dist: precise` explicitly to workaround #1397

yahonda:
It cannot be a permanent fix since Ubuntu Presice is already EOL.
https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming


yahonda:
Refer  #1397

PR #1402, opened 2017-08-02
`remove_prefix_and_suffix` handles dollar sign by `Regexp#escape`

yahonda:
Implemented at Oracle enhanced locally first for rails/rails#30044 Thanks @kamipo


yahonda:
This method will be removed once rails/rails#30048 is merged to Rails master.  Regardless it is merged to Rails Oracle enhanced adapter still needs to support the dollar sign for table_name_prefix and table_name_suffix.

PR #1406, opened 2017-08-03
Use new method name for active record dirty checks

jlahtinen:
This removes deprecation warning inserted in https://github.com/rails/rails/commit/16ae3db5a5c6a08383b974ae6c96faac5b4a3c81

Deprecation warnings are now removed but smees like new methods are more recommendable.

This was problem when doublesaving like described by dhh: https://github.com/rails/rails/commit/16ae3db5a5c6a08383b974ae6c96faac5b4a3c81#commitcomment-20192607

Methods listed
https://github.com/rails/rails/pull/25337#issuecomment-225166796

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/procedures.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/procedures.rb
@@ -146,7 +146,7 @@ module ActiveRecord #:nodoc:
 
               timestamp_attributes_for_update_in_model.each do |column|
                 column = column.to_s
-                next if attribute_changed?(column)
+                next if will_save_change_to_attribute?(column)
                 write_attribute(column, current_time)
               end
             end
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -75,7 +75,7 @@ module ActiveRecord
 
       def record_changed_lobs
         @changed_lob_columns = self.class.lob_columns.select do |col|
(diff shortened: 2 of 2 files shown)


koic:
Thank you for opening PR.
I cannot tell whether the new method names was ultimately decided. Perhaps it seems that warning has disappeared because there is still discussion.

@yahonda I'm interested in your knowledge. What do you think?

yahonda:

Would you add some test cases which represent your changes?

* Showing deprecation messages which will be fixed by this pull request
and/or
* Unexpected behavior fixed by this pull request

I believe rails/rails#25337 "Unresolved questions" still remain because at least none of methods under `ActiveRecord::AttributeMethods::Dirty` are public. Here, I'm not talking about Ruby's public/private/protected ones but Rails public API.

```
Unresolved questions

Do we want the "new behavior" methods to be in the public API at all? They are straight aliases to the existing methods in dirty after 5.2/6.0. However, since we get these bug reports, someone probably does want the new behavior today.
```

Then I do not think this pull request needs to be merged without any actual unexpected behavior fixed by this change.

Thanks.

yahonda:
There is a discussion at https://github.com/rails/rails/issues/30137#issuecomment-321061145

yahonda:
According to the comment above methods under  `ActiveRecord::AttributeMethods::Dirty` will be public. Then I think we can use `will_save_change_to_attribute?` method.

I'm going to merge this pull request but will not backport this change into release18 branch since `will_save_change_to_attribute?` behaves differently with after callbacks.

yahonda:
Merged. Thanks for the pull request.

PR #1407, opened 2017-08-03
Get db user default tablespace and do not rely on USERS

jlahtinen:
If user's default tablespace is not USERS specs does not pass. This asks default tablespace and then asserts it was used to create index.

I am not sure why `def default_tablespace` downcases tablespace name. Now when it does it like that I had to downcase tablespace name used to index also.


yahonda:
Thanks for opening a pull request. However, I do not think we need to merge this once since this `oracle_enhanced` user is for testing Oracle enhanced adapter itself, not for testing Rails applications.

PR #1410, opened 2017-08-08
Fix #1408 check for schema name when validating table name

### LINKED ISSUE #1408: valid_table_name returns false when owner and table name are in different case
### Steps to reproduce
Please refer https://github.com/rsim/oracle-enhanced/tree/master/guides/bug_report_templates
to create an executable test case

https://github.com/rsim/oracle-enhanced/blob/be25c410a71684c912d0135a0d52b31091f574ea/lib/active_record/connection_adapters/oracle_enhanced/quoting.rb#L75

Run this method against 'TEST_DBA.some_table_name'

### Expected behavior
Method should return `true`

### Actual behavior
Method returns `false`

### System configuration
**Rails version**: 4.2.8

**Oracle enhanced adapter version**: 1.6.9

**Ruby version**: 2.4.1

**Oracle Database version**: 11g

It should probably just split on the `.` if one exists and then run the second half of the validation, checking for camelcase, against the table name portion.

Alternatively, this line:
https://github.com/rsim/oracle-enhanced/blob/be25c410a71684c912d0135a0d52b31091f574ea/lib/active_record/connection_adapters/oracle_enhanced/connection.rb#L42

Should just upcase unless the table name is quoted instead of using `valid_table_name?` to decide.

mlh758:
I'm still seeing those 5 failures on my local copy of the repo, but those were failing when I checked the project out from master.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/quoting.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/quoting.rb
@@ -64,8 +64,8 @@ module ActiveRecord
         # and pound sign (#). Database links can also contain periods (.) and
         # "at" signs (@). Oracle strongly discourages you from using $ and # in
         # nonquoted identifiers.
-        NONQUOTED_OBJECT_NAME   = /[A-Za-z][A-z0-9$#]{0,29}/
-        NONQUOTED_DATABASE_LINK = /[A-Za-z][A-z0-9$#\.@]{0,127}/
+        NONQUOTED_OBJECT_NAME   = /[[:alpha:]][\w$#]{0,29}/
+        NONQUOTED_DATABASE_LINK = /[[:alpha:]][\w$#\.@]{0,127}/
         VALID_TABLE_NAME = /\A(?:#{NONQUOTED_OBJECT_NAME}\.)?#{NONQUOTED_OBJECT_NAME}(?:@#{NONQUOTED_DATABASE_LINK})?\Z/
 
         # unescaped table name should start with letter and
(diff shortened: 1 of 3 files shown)


yahonda:
Thanks for opening a pull request.  There are two failures reported with JRuby.  

https://travis-ci.org/rsim/oracle-enhanced/jobs/262307891

Please take a look at the test log and address them. Thanks.

mlh758:
Those tests don't appear to be running when I run the specs locally.

I followed the instructions [here](https://github.com/rsim/oracle-enhanced/blob/master/RUNNING_TESTS.md) and I'm getting 5 failures without any changes to master.

It's going to be difficult to contribute if the specs that run for me are different than the specs that run in Travis.

yahonda:
These two failures only happen when specs run with JRuby.  

Since Oracle enhanced adapter supports both CRuby and JRuby, all spec should pass for both Ruby versions. You can see `master` branch shows all tests are green for both CRuby and JRuby.

yahonda:
Here is the link to Travis CI page for Oracle enhanced adapter. https://travis-ci.org/rsim/oracle-enhanced/branches

mlh758:
I'll look into why the changes aren't working for JRuby. Do you have any insight into why I would be seeing 5 failures on master after following the setup instructions?

mlh758:
I left a comment here detailing what I'm seeing: https://github.com/rsim/oracle-enhanced/issues/1408

mlh758:
I guess it doesn't really matter. I backed out my change in `connection.rb`. I only made it since it seemed weird to check for valid and then continue either way and I was hoping to make the intention clearer. That obviously had unintended side effects in other Ruby versions and it's not worth the extra effort just for a stylistic thing that probably only bothers me.

The tests are all passing now. Do you have any other concerns about my change?

yahonda:
Thanks.

PR #1413, opened 2017-08-14
Setting `disk_asynch_io` to `false` to address slow unit tests

yahonda:
This pull request addresses #1397.
Open a support request with Travis CI and there is a solution provided to set disk_asynch_io=false'

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: precise
+dist: trusty
 sudo: required
 cache: bundler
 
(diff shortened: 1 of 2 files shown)


yahonda:
There are some documents found. I do not know the background yet then FYI.

https://stackoverflow.com/questions/37722685/oracle-11g-docker-select-very-slow
https://forums.docker.com/t/oracle-xe-11g-poor-response-time-in-queries/5209
https://gist.github.com/ihofus/cf697f44f4633c632a12889cfebe7335

PR #1382, opened 2017-07-24
Arity change in insert method

yahonda:
Refer: https://github.com/rails/rails/commit/213796f

This pull request addresses these failures:

```ruby
$ bundle exec rspec ./spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb:218
==> Loading config from ENV or use default
==> Running specs with MRI version 2.4.1
==> Effective ActiveRecord version 5.2.0.alpha
Run options: include {:locations=>{"./spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb"=>[218]}}
F

Failures:

  1) OracleEnhancedAdapter reserved words column quoting should create record
     Failure/Error: super

     ArgumentError:
       wrong number of arguments (given 6, expected 1..5)
     # /home/yahonda/git/rails/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb:140:in `insert'
     # /home/yahonda/git/rails/activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb:19:in `insert'
     # ./lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb:89:in `insert'
     # /home/yahonda/git/rails/activerecord/lib/active_record/relation.rb:64:in `insert'
     # /home/yahonda/git/rails/activerecord/lib/active_record/persistence.rb:595:in `_create_record'
     # /home/yahonda/git/rails/activerecord/lib/active_record/counter_cache.rb:180:in `_create_record'
     # /home/yahonda/git/rails/activerecord/lib/active_record/locking/optimistic.rb:77:in `_create_record'
     # /home/yahonda/git/rails/activerecord/lib/active_record/attribute_methods/dirty.rb:220:in `_create_record'
     # /home/yahonda/git/rails/activerecord/lib/active_record/callbacks.rb:342:in `block in _create_record'
     # /home/yahonda/git/rails/activesupport/lib/active_support/callbacks.rb:132:in `run_callbacks'
     # /home/yahonda/git/rails/activesupport/lib/active_support/callbacks.rb:827:in `_run_create_callbacks'
     # /home/yahonda/git/rails/activerecord/lib/active_record/callbacks.rb:342:in `_create_record'
     # /home/yahonda/git/rails/activerecord/lib/active_record/timestamp.rb:95:in `_create_record'
     # /home/yahonda/git/rails/activerecord/lib/active_record/persistence.rb:569:in `create_or_update'
[...]

PR #1424, opened 2017-08-18
Restore the ability that SQL with binds for `insert`

koic:
Follow up of rails/rails#29944.


yahonda:
Good catch. Refer #1382

koic:
Thanks!

PR #1428, opened 2017-08-21
Support for NCLOB datatype

luke2237:
Added support for NCLOB data types
Updated JDBC connection to handle nclob fields
Added nclob->ntext type to native_database_types.
Updated Specs for NCLOB. They are green.

PR #1440, opened 2017-08-26
Support for NCLOB datatype

luke2237:
Added support for NCLOB datatype. Closed the previous pull request (#1428) and squashed all changes in single commit in this new pull request.


yahonda:
Thank you.

PR #1441, opened 2017-08-26
Move spec files under `oracle_enhanced` directory

yahonda:
This pull request moves spec files under `oracle_enhanced` directory.
#584 introduced `OracleEnhanced` module and created sub directory for modules. It did not touch specs.

PR #1447, opened 2017-08-28
Rename spec/active_record/connection_adapters/oracle_enhanced_dirty_s…

yahonda:
Rename spec/active_record/connection_adapters/oracle_enhanced_dirty_spec.rb to spec/active_record/oracle_enhanced/type/dirty_spec.rb

Pull request #883 removed `lib/active_record/connection_adapters/oracle_enhanced/dirty.rb`
and the logic moved to `changed?` methods under `ActiveRecord::OracleEnhanced::Type`.

It would be too much to split this spec file into each type such as
`string_spec.rb` then this commit just moves `dirty_spec.rb` file to the
corresponding spec directory.

Follow up for #1441


yahonda:
@koic Another review request for you.

koic:
Yes. This is also a dilemma.
I think that it is an ideal to split the spec files into each type. However, moving to under the `oracle_enhanced/type` directory will help us to understand better than now.
Let's start with maintenance on this move. I think that this is a sufficiently practical change.

koic:
Thanks for your idea on this change.

yahonda:

Issue #1449, opened 2017-08-28
Remove nonexistent methods from oracle_enhanced_data_types_spec.rb 


yahonda:
### Steps to reproduce

`oracle_enhanced_data_types_spec.rb`  uses non existent methods.
Since Oracle enhanced adapter 5, Most of these methods were deprecated and removed to
be replaced with Rails attribute API. However, spec names were not changed.

* Refer #963 #871 #872 #874

### Expected behavior
`oracle_enhanced_data_types_spec.rb`  should not have non-existent methods like

* Non existent methods
`emulate_dates`
`emulate_dates_by_column_name`
`emulate_integers_by_column_name`
`set_boolean_columns`
`set_integer_columns`
`set_string_columns`

* Available methods

`emulate_booleans`
`emulate_booleans_from_strings`

### Actual behavior
These methods exist in `oracle_enhanced_data_types_spec.rb` 

### System configuration
**Rails version**:master branch

**Oracle enhanced adapter version**:master branch

**Ruby version**:2.4.1

**Oracle Database version**:Oracle Database 12c Enterprise Edition Release 12.2.0.1.0 - 64bit Production

yahonda:
Closed since these methods have been removed from specs by pull requests linked.

emulate_dates
emulate_dates_by_column_name
emulate_integers_by_column_name
set_boolean_columns
set_integer_columns
set_string_columns

PR #1453, opened 2017-08-28
Remove `set_integer_columns` from `oracle_enhanced_data_types_spec.rb`

yahonda:
Refer #1449


yahonda:
It is not necessary anymore.

PR #1458, opened 2017-08-29
Fixes drop_table causing exception on schema load

BrandonMathis:
## What Changed
* Turn on `if_exists` for the `drop_table` flag in the `create_table` method.

## Why

The command `rake db:schema:load` was raising the following error when it tried to drop a table that did not exist.

```
ActiveRecord::StatementInvalid: NativeException: java.sql.SQLSyntaxErrorException: ORA-00942: table or view does not exist
: DROP TABLE "<TABLE_NAME>" CASCADE CONSTRAINTS
```

This change utilizes the `if_exists` option for the `drop_table` method to prevent raising of this error when schema load goes through the process of dropping tables.

## Note
It would be amazing if this could be deployed to 1.8 and 1.7 since I require 1.7 for use with my Rails 5.0 application.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -72,7 +72,7 @@ module ActiveRecord
           create_sequence = create_sequence || td.create_sequence
 
           if options[:force] && data_source_exists?(table_name)
-            drop_table(table_name, options)
+            drop_table(table_name, options.merge({ if_exists: true }))
           end
 
           execute schema_creation.accept td
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -72,7 +72,7 @@ module ActiveRecord
           create_sequence = create_sequence || td.create_sequence
 
           if options[:force] && data_source_exists?(table_name)
-            drop_table(table_name, options.merge({ if_exists: true }))
+            drop_table(table_name, options.merge(if_exists: true))
           end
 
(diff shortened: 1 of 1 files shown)


yahonda:

Here `drop_table` is executed only when `if options[:force] && data_source_exists?(table_name)`  is satisfied. So adding `if_exists: true` should not be necessary. 

I'd like to know why `drop_table` method called here. Would you provide steps to reproduce this error.

Thanks.

BrandonMathis:
Interesting. [Rails' schema dumper now sets `cascade: true` on all tables](http://edgeguides.rubyonrails.org/4_2_release_notes.html#active-record-notable-changes) so it seems that this condition will always be fulfilled.

I guess [`data_source_exists?`](https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L561-L563) is the root of the issue since it should evaluate to false since the table is not present (hence the exception).

It is likely worth noting that I am using jruby 9.1.12.0

Reproduction steps should be:  
1. Create new rails app
1. Install enhanced adapter
1. Create migration that makes some db table
1. `rake db:migrate`
1. `rake db:schema:load`

BrandonMathis:
Interesting. Maybe I am connecting to a user that has that table already defined and it's causing some trouble?

Will investigate further and report back 👍

BrandonMathis:
I was specifically experiencing this issue when running `rake db:schema:load`. When the task got to the point of creating a 'user_credentials' table for it, this error happened.

I've come to the conclusion that this table name must already be used by someone else or oracle itself. Not 100% sure but I managed to get around this error by renaming the table.

Closing PR

yahonda:
Thanks for the update.

PR #1459, opened 2017-08-29
Extract JSON data type specs

yahonda:
into `spec/active_record/oracle_enhanced/type/json_type_spec.rb`


yahonda:
This is the first step to extract specs for each type from `oracle_enhanced_data_types_spec.rb`.

Let me explain the background and the mid term goal.

Originally, `oracle_enhanced_data_types_spec.rb` was testing Oracle database schema created outside of Rails migration.  These `emulate_dates` `emulate_dates_by_column_name` `emulate_integers_by_column_name` `set_boolean_columns` `set_integer_columns` `set_string_columns` were created to support these non-Rails style schema. Therefore, some of tables are created by `create table` sql statements, not by `create_table` method.

Starting from Rails 5, these methods were removed and replaced by Rails 5 attribute API support.
Most of type specic implementations moved to each type under `ActiveRecord::OracleEnhanced::Type`.

Once this pull request merged, extract specs for each type, such as `integer_type_spec.rb`, `boolean_type_spec.rb` will be created like this. Each spec will test Rails attribute API support. Finally,  `oracle_enhanced_data_types_spec.rb` will be removed after these extraction completed.

yahonda:
Merging this pull request. 

@koic I think you would agree with this change but feel free to give some comments for this pull request and mid-term goal to remove `oracle_enhanced_data_types_spec.rb` above.

koic:
@yahonda Thanks for improvement activity. This change is just how I had imagined it.
There is one feedback in that.

If I had to choose, `spec/active_record/oracle_enhanced/type/json_spec.rb` rather than `spec/active_record/oracle_enhanced/type/json_type_spec.rb` may be better. Because it is the testing pair of `lib/active_record/oracle_enhanced/type/json.rb`. And since `active_record/oracle_enhanced/type` directory points to `type`, there is no redundancy.

I think that #1460, #1461, and #1462 are also `type/*_spec.rb` is better than `type/*_type_spec.rb`.

koic:
I opened a PR #1463 that extracted `integer_spec.rb` as an example.

yahonda:
#1463 looks good to me. Will open pull requests to change the location of test files to follow the new convention.

Issue #1467, opened 2017-08-30
Remaining 5 descibes at oracle_enhanced_data_types_spec.rb


yahonda:
### Steps to reproduce

Many of data type specs have been extracted 
from `oracle-enhanced/spec/active_record/connection_adapters/oracle_enhanced_data_types_spec.rb`
to `oracle-enhanced/spec/active_record/oracle_enhanced/type/*_spec.rb`  Refer #1459 #1460 #1461 #1462 #1463 #1464 #1465 #1466

All of them have been straightforward so far because there are Oracle enhanced adapter own types corresponding to each spec. 

As of now `oracle_enhanced_data_types_spec.rb` has 5 top level "describe"s below:

* 5 top level describes in `oracle_enhanced_data_types_spec`
```ruby
describe "OracleEnhancedAdapter date and datetime type detection based on attribute settings" do
describe "OracleEnhancedAdapter timestamp with timezone support" do
describe "OracleEnhancedAdapter assign string to :date and :datetime columns" do
describe "OracleEnhancedAdapter handling of BLOB columns" do
describe "OracleEnhancedAdapter handling of BINARY_FLOAT columns" do
```

The extraction will not be straightforward for remaining describes.  I'll explain them each.

1. describe "OracleEnhancedAdapter date and datetime type detection based on attribute settings" do

- This describe tests both `date` and `datetime` type. 
- Oracle enhanced adapter does not implement its own types.

2. describe "OracleEnhancedAdapter timestamp with timezone support" do
- This describe tests `timestamp` `timestamptz` and `timestamp_ltz` type. 
- Oracle enhanced adapter implements both `timestamptz` and `timestamp_ltz` type. Use abstract implementation for `timestamp`
- I prefer to keep this describe unit. Do not want to test each `timestamp` `timestamptz` and `timestamp_ltz` type separately.

3. describe "OracleEnhancedAdapter assign string to :date and :datetime columns" do
- This describe tests both `date` and `datetime` type. 
- Oracle enhanced adapter does not implement its own types.

4. describe "OracleEnhancedAdapter handling of BLOB columns" do
- Oracle enhanced adapter does not implement its own `binary` type.

5. describe "OracleEnhancedAdapter handling of BINARY_FLOAT columns" do
- Oracle enhanced adapter does not implement its own `float` type.

### Expected behavior
In the long term, oracle_enhanced_data_types_spec.rb file is removed.

### Actual behavior
[...]

PR #1480, opened 2017-09-09
Extract TIMPSTAMP, TIMESTAMPTZ and TIMESTAMPLTZ specs

yahonda:
Refer #1467 

> describe "OracleEnhancedAdapter timestamp with timezone support" do
...
> I prefer to keep this describe unit. Do not want to test each timestamp timestamptz and timestamp_ltz type separately.


koic:
This PR summarizes as `timestamp` tests that `created_at_tz` and `created_at_ltz` work correctly for `Type::TimestampTz` and `Type::TimestampLtz` type classes.

It might be better to make these files as separate files, but because of the disadvantage of becoming too small, let's first try this way of splitting.

PR #1481, opened 2017-09-09
Extract BINARY type specs

yahonda:
Refer #1467

> describe "OracleEnhancedAdapter handling of BLOB columns" do
> Oracle enhanced adapter does not implement its own binary type.

It would need some discussion, I prefer to move these specs to under
type.


koic:
Same as https://github.com/rsim/oracle-enhanced/pull/1482#pullrequestreview-62371931.

koic:
@yahonda Would you resolve the conflict and merge it?

yahonda:
Conflict resolved and merged. Thanks for the discussion and reviews for this and other two pull requests.

PR #1482, opened 2017-09-09
Extract FLOAT type specs

yahonda:
Refer #1467

> describe "OracleEnhancedAdapter handling of BINARY_FLOAT columns" do
> Oracle enhanced adapter does not implement its own float type.

It also needs some more discussions.


koic:
This may be a test for quoting, but first of all I will merge it as a step of splitting from a fat `oracle_enhanced_data_types_spec.rb`.

PR #1488, opened 2017-09-17
Oracle enhanced adapter version follows Rails versioning

yahonda:
* Major X and Minor Y - They will follow Rails versions
* Patch Z - Oracle enhanced adapter has its own patch versions

http://guides.rubyonrails.org/maintenance_policy.html


yahonda:
Refer this announcement:
https://groups.google.com/forum/#!topic/oracle-enhanced/ZevW3CD8vC0

koic:
This is easy to understand the versioning relationship with Rails. I prefer this versioning.

PR #1508, opened 2017-10-02
Remove `value_to_boolean` which has been removed from Rails

yahonda:
Remove `value_to_boolean` which has been removed from Rails since 4.2

Refer https://github.com/rails/rails/commit/8f387995054d9ce79522da8ef65fba28c6ea0785

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/column.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/column.rb
@@ -40,18 +40,6 @@ module ActiveRecord
           @object_type
         end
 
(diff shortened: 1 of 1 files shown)


awaltman:
Will this impact the translation of Y/N values to true/false? Do you know if the rails maintainers are replacing this functionality with something else?

Thanks!

awaltman:
I see now how they are inlining the methods. However, at first glance it does look like this will break converting Y and N to true and false in oracle-enhanced since Column::TRUE_VALUES does not contain 'Y' or 'y'.

PR #1526, opened 2017-10-09
Let `ActiveRecord::ConnectionAdapters::OracleEnhanced::TypeMetadata` handle `virtual` type

yahonda:
There have been some refactoring `ActiveRecord::ConnectionAdapters::OracleEnhanced::Column` made. Now refactoring for  `virtual` type support remains.

This pull request is work in progress one. The main problem so far is `fetch_type_metadata` does not return ` ActiveRecord::ConnectionAdapters::OracleEnhanced::TypeMetadata` including `virtual` then it cannot handle `virtual` which used to be inside `ActiveRecord::ConnectionAdapters::OracleEnhanced::Column`.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -482,19 +482,20 @@ module ActiveRecord
               field["data_default"] = false if (field["data_default"] == "N" && OracleEnhancedAdapter.emulate_booleans_from_strings)
             end
 
-            type_metadata = fetch_type_metadata(field["sql_type"])
+            type_metadata = fetch_type_metadata(field["sql_type"], is_virtual)
+            default_value = extract_value_from_default(field["data_default"])
+            default_value = nil if is_virtual
             OracleEnhanced::Column.new(oracle_downcase(field["name"]),
-                             field["data_default"],
+                             default_value,
                              type_metadata,
                              field["nullable"] == "Y",
                              table_name,
-                             is_virtual,
                              field["column_comment"]
                       )
           end
 
-          def fetch_type_metadata(sql_type)
-            OracleEnhanced::TypeMetadata.new(super(sql_type))
+          def fetch_type_metadata(sql_type, virtual = nil)
+            OracleEnhanced::TypeMetadata.new(super(sql_type), virtual: virtual)
           end
 
           def tablespace_for(obj_type, tablespace_option, table_name = nil, column_name = nil)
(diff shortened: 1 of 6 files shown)


yahonda:
@kamipo Would you do me a favor?  I'm struggling with `fetch_type_metadata`. It does not return ` ActiveRecord::ConnectionAdapters::OracleEnhanced::TypeMetadata` including `virtual`  then it cannot handle `virtual` which used to be inside `ActiveRecord::ConnectionAdapters::OracleEnhanced::Column`.

yahonda:
Here are steps how to create Oracle enhanced adapter development environment with macOS https://gist.github.com/yahonda/d4936bd065db9aa421f3ea95b5473a2e

lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
```diff
@@ -482,19 +482,23 @@ def new_column_from_field(table_name, field)
               field["data_default"] = false if (field["data_default"] == "N" && OracleEnhancedAdapter.emulate_booleans_from_strings)
             end
 
-            type_metadata = fetch_type_metadata(field["sql_type"])
+            type_metadata = fetch_type_metadata(field["sql_type"], is_virtual)
+            # Somehow, `type_metadata` does not have `virtual` and its class is `ActiveRecord::ConnectionAdapters::SqlTypeMetadata`
+            # Not ActiveRecord::ConnectionAdapters::OracleEnhanced::TypeMetadata`
+            require "pp"
+            pp type_metadata if is_virtual
```
kamipo: Can you confirm `p type_metadata.class`?
`pp` calls `object.pretty_print`, so `DelegateClass` will delegate it to the instance of `SqlTypeMetadata` in `OracleEnhanced::TypeMetadata`.

yahonda: Thanks for the answer.

`p type_metadata.class if is_virtual` shows
`ActiveRecord::ConnectionAdapters::OracleEnhanced::TypeMetadata` as expected, `p type_metadata.inspec` shows `"#<ActiveRecord::ConnectionAdapters::SqlTypeMetadata:0x0000000003a1f8f8 @sql_type=\"VARCHAR2(512)\", @type=:string, @limit=512, @precision=nil, @scale=nil>"`, I do not know if `object#inspect` is affected by delegate or not but I wanted it shows `virtual`
Maybe something wrong with `fetch_type_metadata` method itself.  Will investigate.

yahonda: I think `fetch_type_metadata` is working fine now. All specs are green.

PR #1540, opened 2017-10-13
Suppress `DEPRECATION WARNING: Dangerous query method`

yahonda:
This pull request has been created for some more discussion in advance in case rails/rails/#27947 merged to master.

It is easy to suppress `DEPRECATION WARNING: Dangerous query method` by adding `Arel.sql` but I am not sure if it is a better way to implement it.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/context_index.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/context_index.rb
@@ -332,7 +332,7 @@ module ActiveRecord
             def contains(column, query, options = {})
               score_label = options[:label].to_i || 1
               where("CONTAINS(#{connection.quote_table_name(column)}, ?, #{score_label}) > 0", query).
-                order("SCORE(#{score_label}) DESC")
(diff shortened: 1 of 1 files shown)


yahonda:
Refer rails/rails#27947

yahonda:
Paste the commit message of the first commit in this pull request:

Suppress `DEPRECATION WARNING: Dangerous query method`

```ruby
DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): "SCORE(0) DESC". Non-attribute arguments will be disallowed in Rails 6.0. This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql(). (called from contains at /path/to//oracle-enhanced/lib/active_record/connection_adapters/oracle_enhanced/context_index.rb:335)
```
Support rails/rails#27947

May need to consider how to address in the long term since `SCORE` is not a standard SQL literal.

yahonda:
Changed the subject since  rails/rails#27947 has been merged to master.

PR #1541, opened 2017-10-16
Rename `ActiveRecord::OracleEnhanced::Type` to `ActiveRecord::Type::OracleEnhanced`

yahonda:
Rename `ActiveRecord::OracleEnhanced::Type` to `ActiveRecord::Type::OracleEnhanced`
and use relative path to keep the same structure with Rails and make code shorter.

Except for specs, which would get this error:

```ruby
     Failure/Error: binds = [ActiveRecord::Relation::QueryAttribute.new("id", 1, Type::OracleEnhanced::Integer.new)]

     NameError:
       uninitialized constant Type
```


yahonda:
@koic Would you take a look at this pull request and review if you agree with this change?

koic:
I think that it is a good namespace change to move to `ActiveRecord::Type` as a place for an OracleEnhanced type classes that inherits `ActiveModel::Type`.

yahonda:
Thanks for the review.

PR #1547, opened 2017-10-20
Introduce `OracleEnhanced::Lob` not to modify `ActiveRecord::Base` directly

yahonda:
To load LOB data from fixtures, Oracle enhanced adapter
has modified `ActiveRecore::Base`, which can be avoided
by introducing `OracleEnhanced::Lob` module using `ActiveSupport::Concern`

TODO:
- [x]  Changed `ActiveRecord::Base` to somewhat appropriate
like `ActiveRecord::ConnectionAdapters::OracleEnhanced::Lob` module

- [x]  Better way to use the new module such as with `ActiveSupport::Concern`
Refer http://api.rubyonrails.org/classes/ActiveSupport/Concern.html

- [x]  `ApplicationRecord` may not be an appropriate way since
this will be used for `insert_fixtures` which is defined in the Rails itself.

- [x] Rename `base.rb` to `lob.rb`

- [x] Where and how to handle `ActiveRecord::Base.class_eval`

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/lob.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/lob.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module ActiveRecord #:nodoc:
+  module ConnectionAdapters #:nodoc:
+    module OracleEnhanced #:nodoc:
+      module Lob #:nodoc:
+        extend ActiveSupport::Concern
+
+        included do
+          class_attribute :custom_create_method, :custom_update_method, :custom_delete_method
+
+          # After setting large objects to empty, select the OCI8::LOB
+          # and write back the data.
+          before_update :record_changed_lobs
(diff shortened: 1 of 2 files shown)


yahonda:
@koic Would you give me some comments about this change?

koic:
Looks great! About the handling of LOB It seems to be clarified by extracting to the module. The big file `oracle_enhanced_adapter.rb` will also be slightly smaller.
There may be a more appropriate place to mix-in `OracleEnhanced::Lob` module into `ActiveRecord::Base` class, but first of all I think that this change is good.

yahonda:

> There may be a more appropriate place to mix-in OracleEnhanced::Lob module into ActiveRecord::Base class, 

Agreed. but I do not have other ideas. if you have some better fix, please open a pull request.

Thanks again for the review.

koic:
Noted 📝 . I don't have a good idea yet, but I will open a PR if I get a good idea. Thanks always.

PR #1550, opened 2017-10-23
Log the purpose of sql in `disable_referential_integrity`

yahonda:
* With this commit

```ruby
  Foreign Keys to disable and enable (33.7ms)  SELECT constraint_name, owner, table_name FROM all_constraints WHERE constraint_type = 'R' AND status = 'ENABLED' AND owner = SYS_CONTEXT('userenv', 'session_user')
```

* Without this commit

```ruby
   (35.1ms)            SELECT constraint_name, owner, table_name
            FROM all_constraints
            WHERE constraint_type = 'R'
            AND status = 'ENABLED'
            AND owner = SYS_CONTEXT('userenv', 'session_user')
```

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -411,14 +411,13 @@ module ActiveRecord
         # REFERENTIAL INTEGRITY ====================================
 
         def disable_referential_integrity(&block) #:nodoc:
-          sql_constraints = <<-SQL
+          old_constraints = select_all(<<-SQL.strip.gsub(/\s+/, " "), "Foreign Keys to disable and enable")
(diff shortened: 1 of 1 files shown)


yahonda:
@koic Would you tell me which log output do you prefer, with or without this commit?

* Having the necessary description of the SQL like "Foreign Keys to disable and enable"
* One line SQL statement even if the code has several CRs

koic:
I agree with this change.

As I looked at the log of AR with Oracle enhanced adapter, it is preferable to make internal SQL one liner. Because AR and Oracle enhanced adapter's internal SQL output is unified into one liner, then log analysis becomes easy. 
Next, as for description of the SQL, information will be added like "Primary Key", "Sequence" and others I think that it increase easy to trace (It is preferable if there is one).

yahonda:
Thanks for the comment.

koic:
Thanks too!

PR #1553, opened 2017-10-24
Log multi lines SQL statements into one line

yahonda:
and add necessary notes for each `select_one`, `select_value`, `select_values`, `select_all` method call.

Note: this commit does not change `structure_dump.rb`. it looks complicated.


yahonda:
Refer #1550 and #1551

PR #1555, opened 2017-10-24
Log multi lines SQL statements into one line

yahonda:
and add necessary notes at structure_dump.rb

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/structure_dump.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/structure_dump.rb
@@ -8,23 +8,36 @@ module ActiveRecord #:nodoc:
         STATEMENT_TOKEN = "\n\n/\n\n"
 
         def structure_dump #:nodoc:
-          structure = select("SELECT sequence_name, min_value, max_value, increment_by, order_flag, cycle_flag FROM all_sequences where sequence_owner = SYS_CONTEXT('userenv', 'session_user') ORDER BY 1").map do |result|
+          sequences = select(<<-SQL.strip.gsub(/\s+/, " "), "sequences to dump at structure dump")
+            SELECT sequence_name, min_value, max_value, increment_by, order_flag, cycle_flag
+            FROM all_sequences
+            where sequence_owner = SYS_CONTEXT('userenv', 'session_user') ORDER BY 1
+          SQL
+
+          structure = sequences.map do |result|
             "CREATE SEQUENCE #{quote_table_name(result["sequence_name"])} MINVALUE #{result["min_value"]} MAXVALUE #{result["max_value"]} INCREMENT BY #{result["increment_by"]} #{result["order_flag"] == 'Y' ? "ORDER" : "NOORDER"} #{result["cycle_flag"] == 'Y' ? "CYCLE" : "NOCYCLE"}"
           end
-          select_values("SELECT table_name FROM all_tables t
-                      WHERE owner = SYS_CONTEXT('userenv', 'current_schema') AND secondary = 'N'
-                        AND NOT EXISTS (SELECT mv.mview_name FROM all_mviews mv WHERE mv.owner = t.owner AND mv.mview_name = t.table_name)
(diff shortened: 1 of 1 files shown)


koic:
I came up with "Method Chain" and "Extract Method". The following is an images of the change.

## Method Chain

```diff
          tables.each do |table_name|
            virtual_columns = virtual_columns_for(table_name) if supports_virtual_columns?
            ddl = "CREATE#{ ' GLOBAL TEMPORARY' if temporary_table?(table_name)} TABLE \"#{table_name}\" (\n".dup
            columns = select_all(<<-SQL.strip.gsub(/\s+/, " "), "columns at structure dump")
              SELECT column_name, data_type, data_length, char_used, char_length,
              data_precision, data_scale, data_default, nullable
              FROM all_tab_columns
              WHERE table_name = '#{table_name}'
              AND owner = SYS_CONTEXT('userenv', 'session_user')
              ORDER BY column_id
            SQL
-           cols = columns.map do |row|
-             if (v = virtual_columns.find { |col| col["column_name"] == row["column_name"] })
-               structure_dump_virtual_column(row, v["data_default"])
-             else
-               structure_dump_column(row)
-             end
-           end
-           ddl << cols.map { |col| " #{col}" }.join(",\n")
+           ddl << columns.map { |row|
+             if (v = virtual_columns.find { |col| col["column_name"] == row["column_name"] })
+               structure_dump_virtual_column(row, v["data_default"])
+             else
+               structure_dump_column(row)
+             end
+           }.map { |col| " #{col}" }.join(",\n")
            ddl << structure_dump_primary_key(table_name)
```

## Extract Method

```diff
           tables.each do |table_name|
             virtual_columns = virtual_columns_for(table_name) if supports_virtual_columns?
             ddl = "CREATE#{ ' GLOBAL TEMPORARY' if temporary_table?(table_name)} TABLE \"#{table_name}\" (\n".dup
-            columns = select_all(<<-SQL.strip.gsub(/\s+/, " "), "columns at structure dump")
-              SELECT column_name, data_type, data_length, char_used, char_length,
-              data_precision, data_scale, data_default, nullable
-              FROM all_tab_columns
-              WHERE table_name = '#{table_name}'
-              AND owner = SYS_CONTEXT('userenv', 'session_user')
-              ORDER BY column_id
-            SQL
-            cols = columns.map do |row|
-              if (v = virtual_columns.find { |col| col["column_name"] == row["column_name"] })
-                structure_dump_virtual_column(row, v["data_default"])
-              else
-                structure_dump_column(row)
-              end
-            end
-            ddl << cols.map { |col| " #{col}" }.join(",\n")
+            ddl << structure_dump_columns(table_name)
             ddl << structure_dump_primary_key(table_name)

...

+         def structure_dump_columns(table_name)
+           columns = select_all(<<-SQL.strip.gsub(/\s+/, " "), "columns at structure dump")
+             SELECT column_name, data_type, data_length, char_used, char_length,
+             data_precision, data_scale, data_default, nullable
+             FROM all_tab_columns
+             WHERE table_name = '#{table_name}'
+             AND owner = SYS_CONTEXT('userenv', 'session_user')
+             ORDER BY column_id
+           SQL
+           columns.map do |row|
+             if (v = virtual_columns.find { |col| col["column_name"] == row["column_name"] })
+               structure_dump_virtual_column(row, v["data_default"])
+             else
+               structure_dump_column(row)
+             end
+           end
+         end
```

I think that "Extract Method" is better than "Method Chain" for the following 2 reasons.

- `structure_dump` method has a large number of rows.
- I see some internal methods called `structure_dump_xxx`.

However, because it is not superior to the current code 💦 Since current PR code may be better, I will Approve.

yahonda:
Thanks for the detailed examples.
I intended something like "Method Chain" pattern.

Considering some more refactoring by using "Extract Method" pattern,
I'd rather prefer to consider to implement other tools which generate schema information by SQL statements. Since PostgreSQL structure dump actually calls `pg_dump`, MySQL one calls `mysqldump`.
Unfortunately, Oracle native expdp/impdp is not such simple. Then I'll keep the current implementation of structure dump.

Thanks again for the review and agreed. Let's merge this pull request.

PR #1561, opened 2017-10-26
Remove `OracleEnhanced::Connection#select_value`

yahonda:
Since #1560 it is not called from Oracle enhanced adapter specs.

Usually `ActiveRecord::ConnectionAdapters::DatabaseStatements#select_value`
should be used.


yahonda:
Need to address spec failures with JRuby.

PR #1563, opened 2017-10-27
Rename `OracleEnhanced::Connection#select_value` to `_select_value`

yahonda:
Rename `OracleEnhanced::Connection#select_value` to `_select_value` and `OracleEnhanced::Connection#select_one` to `_select_one`
and make them private to avoid other methods all them.

Since #1560 they are only called from `OracleEnhanced::Connection#describe`

Usually `ActiveRecord::ConnectionAdapters::DatabaseStatements#select_value`
and `ActiveRecord::ConnectionAdapters::DatabaseStatements#select_one` should be used.

`OracleEnhanced::Connection#describe` is planned to refactor in the next
major version of Oracle enhanced adapter, likely 6.0 or later.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
@@ -34,7 +34,7 @@ module ActiveRecord
           name = name.to_s
           if name.include?("@")
             name, db_link = name.split("@")
-            default_owner = select_value("SELECT username FROM all_db_links WHERE db_link = '#{db_link.upcase}'")
+            default_owner = _select_value("SELECT username FROM all_db_links WHERE db_link = '#{db_link.upcase}'")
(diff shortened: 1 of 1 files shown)


yahonda:
@koic Would you provide some comments and reviews for this pull request?

yahonda:
Background for this pull request is also explained in #1560 #1559 #1558 .

koic:
I don't have any problems to worry about this change and background PRs. Is there anything you care about this change?

yahonda:
Thanks for the review. 

The reason why I asked a review was adding a _ to each method was an appropriate way to handle this. 
I was thinking about adding a deprecated warning for these two methods but I prefer these methods will not be executed by adding _ . I should have mentioned this. Will include the reason for future review requests.

Thanks.

koic:
I got it. This looks good. Thanks!

PR #1564, opened 2017-10-27
Modularize `OracleEnhanced::Connection` class

yahonda:
Modularize `ActiveRecord::ConnectionAdapters::OracleEnhanced::Connection` class

`OracleEnhanced::Connection` class itself does not create connections by itself.
It mainly just handles if ruby runtime is CRuby or JRuby then
creates `OracleEnhanced::OCIConnection` or `OracleEnhanced::JDBCConnection` instances.

Also, there is almost unnecessary to create `Connection#database_version`.
Just let each `OracleEnhanced::OCIConnection` and `OracleEnhanced::JDBCConnection`
class includes `OracleEnhanced::Connection` module and implement
their own `database_version` methods.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
@@ -4,7 +4,7 @@ module ActiveRecord
   module ConnectionAdapters
     # interface independent methods
     module OracleEnhanced
-      class Connection #:nodoc:
+      module Connection #:nodoc:
         def self.create(config)
           case ORACLE_ENHANCED_CONNECTION
           when :oci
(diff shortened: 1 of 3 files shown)


yahonda:
@koic Another review request for you.

koic:
The name `Connection` seems to be a class. For example, the name `Connectable` might look like a module.

It is also a different view.
Although it is not a strong opinion. I don't feel a problem with the current implementation. This is because `OCIConnection` class and `JDBCConnection` class inherit `OracleEnhanced::Connection` class obviously is an "is-a" relationship.
On the other hand, if there are no additional connections beyond OCI and JDBC in the future, remove `database_version` method from `OracleEnhanced::Connection` class will be fine.

yahonda:
Thanks for the review and agreed. 

>  This is because OCIConnection class and JDBCConnection class inherit OracleEnhanced::Connection class obviously is an "is-a" relationship.

That really makes sense. Will close this pull request.

PR #1565, opened 2017-10-30
Rename `OracleEnhanced::Connection#oracle_downcase` to `_oracle_downc…

yahonda:
Rename `OracleEnhanced::Connection#oracle_downcase` to `_oracle_downcase`

`_oracle_downcase` is expected to be called only from `ActiveRecord::ConnectionAdapters::OracleEnhanced::OCIConnection` or `ActiveRecord::ConnectionAdapters::OracleEnhanced::JDBCConnection`.

Other method should call `ActiveRecord:: ConnectionAdapters::OracleEnhanced::Quoting#oracle_downcase`
since this is kind of quoting, not connection.
To avoid it is called from anywhere else, added _ at the beginning of the method name.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
@@ -18,17 +18,6 @@ module ActiveRecord
 
         attr_reader :raw_connection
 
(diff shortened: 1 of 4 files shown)


yahonda:
Refer #1563 for the same kind of changes.

PR #1566, opened 2017-10-30
Make `OracleEnhanced::Connection#describe` private

yahonda:
which is only called from its subclasses of `OracleEnhanced::Connection`

`describe` method may be appropriate to move to `OracleEnhanced::SchemaStatements`
which has `table_exists?` or similar methods.

but I'd like to postpone this decision. Just make it private now.


yahonda:
The biggest reason why I postpone this decision is I'm inclined to remove `OracleEnhanced::Connection#describe` in the future version of Oracle enhanced adapter.  By moving these logic into `tables` , or `views`.

PR #1608, opened 2017-12-15
Use latest Rubies on Travis CI

koic:
These Rubies has been released.

- https://www.ruby-lang.org/en/news/2017/12/14/ruby-2-2-9-released/
- https://www.ruby-lang.org/en/news/2017/12/14/ruby-2-3-6-released/
- https://www.ruby-lang.org/en/news/2017/12/14/ruby-2-4-3-released/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,9 @@ install:
 
 language: ruby
 rvm:
-  - 2.2.8
-  - 2.3.5
-  - 2.4.2
+  - 2.2.9
+  - 2.3.6
+  - 2.4.3
   - ruby-head
   - jruby-9.1.15.0
   - jruby-head


yahonda:
Thanks. would you back port this change to `release-18` branch?

Also, I'm wondering if we need to test with Ruby 2.5.0 rc1
https://www.ruby-lang.org/en/news/2017/12/14/ruby-2-5-0-rc1-released/

As long as Ruby 2.5.0 is going to be released from the same branch as we use `ruby-head` it is fine to use `ruby-head` until Ruby 2.5.0 is released. but not sure.

koic:
Thanks for your comment. I'm going to backport to the `release-18` branch (I was forgotten about it).

> As long as Ruby 2.5.0 is going to be released from the same branch as we use ruby-head it is fine to use ruby-head until Ruby 2.5.0 is released.

I agree.
Ruby 2.5.0-rc1 is a preview snapshot, so it would be fine to use ruby-head.

PR #1613, opened 2017-12-24
Ruby 2.5 does not have bundler as bundled gem

yahonda:
Also applied https://github.com/rails/rails/pull/31541

Kind of reverting #1532


yahonda:
Close until #1614 resolved.

Issue #1611, opened 2017-12-22
Time column behaviour is different from other versions (1.6)


jameschenjav:
### Steps to reproduce
1. Model `Foo` contains a timestamp column `bar`.
2. Get an instance: `foo = Foo.first`
3. Assign a string to the timestamp column `foo.bar = '2017-12-22 01:02:03'`

### Expected behavior
The type of `foo.bar` should be `Time` or similar types like `ActiveSupport::TimeWithZone`

### Actual behavior
It's `String` in 1.6
I tested it in 1.5~1.8. All the other versions are correct.

### System configuration
**Rails version**:
4.2.10

**Oracle enhanced adapter version**:
1.6.9 (from Gemfile.lock)

**Ruby version**:
2.4.2

**Oracle Database version**:
11g release 11.2.0.1.0

yahonda:
Hi, Would you create a executable test case based on `active_record_gem.rb` or `active_record_gem_spec.rb`  at https://github.com/rsim/oracle-enhanced/tree/master/guides/bug_report_templates

Thank you.

eGust:
would these be ok?

[bug_report_templates.zip](https://github.com/rsim/oracle-enhanced/files/1592407/bug_report_templates.zip)

fail only
```ruby
  gem "rails", github: "rails/rails", branch: "4-2-stable"
  gem "arel" #, github: "rails/arel", branch: "master"
  gem "activerecord-oracle_enhanced-adapter",  github: "rsim/oracle-enhanced", branch: "release16"
```

```
  1) bug test should be a Time type after assigning
     Failure/Error: expect(post.created_at.class.name).to eq("Time") # both Time and ActiveSupport::TimeWithZone use "Time"

       expected: "Time"
            got: "String"

       (compared using ==)
     # ./active_record_gem_spec.rb:65:in `block (2 levels) in <top (required)>'
```

### active_record_gem.rb
```ruby
ActiveRecord::Schema.define do
  create_table :posts, force: true do |t|
    t.timestamp :created_at
  end
end

class Post < ActiveRecord::Base
end

class BugTest < Minitest::Test
  def test_timestamp_assignment
    post = Post.new
    post.created_at = "2000-01-02 03:04:05"
    assert_equal post.created_at.class.name, "Time" # both Time and ActiveSupport::TimeWithZone
  end
end
```

### active_record_gem_spec.rb
```ruby
describe "bug test" do
  before(:all) do
[...]

PR #1620, opened 2018-01-04
#1611 OracleEnhancedAdapter 1.6.X timestamp cast bugfix

bardbess:
When assigning a string date to a timestamp column it is not being converted to a Time instance - issue #1611

- New test in data_types_spec for timestamp column string conversion
- Fix on OracleEnhanced Timestamp type attribute cast for Rails 4.2

### DIFF
--- a/lib/active_record/oracle_enhanced/type/timestamp.rb
+++ b/lib/active_record/oracle_enhanced/type/timestamp.rb
@@ -5,6 +5,14 @@ module ActiveRecord
         def type
           :timestamp
         end
+
+        def type_cast_from_user(value)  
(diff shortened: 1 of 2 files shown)


yahonda:
Thanks for opening a pull request. Unfortunately as I mentioned in https://github.com/rsim/oracle-enhanced/issues/1611#issuecomment-355316107 Oracle enhanced adapter 1.6.x is not maintained anymore then this pull request would not be merged. 

Thanks again for the investigation and opening a pull request.

yahonda:
Closing this pull request since no change will be made for `release16` branch.

PR #1627, opened 2018-01-16
Support schema names with only numbers

josephpage:
When a schema name is only numbers, eg. "0000" or "1", all request failed with this error : 
```
OCIError: ORA-02421: missing or invalid schema authorization identifier
```
In this PR, I force the SQL `alter session` to understand the name as a string.
Thanks for reviewing !

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
@@ -168,7 +168,7 @@ module ActiveRecord
             # default schema owner
             @owner = username.upcase unless username.nil?
           else
-            exec "alter session set current_schema = #{schema}"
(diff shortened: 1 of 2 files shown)


yahonda:
Hi, Thanks for opening a pull request.

Database user name should be nonquoted identifier, begining with an alphabetic character.

Database user name only names is a quoted identifier,which is not recommended.

I'd suggest you to change your policy to follow Oracle recommendation. If you still would like to use database users with numbers, you can modify your own code by youself since this pull request shows some failures with RSpec. but I prefer not to merge for other users.


* Refer these documents what is "quoted identifier", "nonquoted identifier"

https://docs.oracle.com/database/121/SQLRF/statements_8003.htm#SQLRF01503


https://docs.oracle.com/database/121/SQLRF/sql_elements008.htm#SQLRF00223


Thanks.

PR #1630, opened 2018-01-19
Enable autocorrect for `Lint/EndAlignment` cop

yahonda:
Refer rails/rails#31732


koic:
This PR looks the same contents as #1629. Is there any intention in this PR?

Issue #1595, opened 2017-11-20
Change `default_sequence_start_value` from 10000 to 1


yahonda:
### Steps to reproduce
Create a Rails application

### Expected behavior
`id` value starts from `1` as other ActiveRecord bundled adapters do.
If users who wanted to keep the current behavior, they can use ` ActiveRecord::ConnectionAdapters::OracleEnhancedAdapter.default_sequence_start_value = 10000`

### Actual behavior
`id` value starts from `10000`, which can be configured ` ActiveRecord::ConnectionAdapters::OracleEnhancedAdapter.default_sequence_start_value = 1`

### System configuration
**Rails version**:next major

**Oracle enhanced adapter version**:next major

**Ruby version**: Any ruby version supported by the next major version of Rails

**Oracle Database version**:11.2 or higher

yahonda:
Was looking for  the reason why the start value is 10000 but unable to find it.

* This value has been set from the first commit
https://github.com/rsim/oracle-enhanced/commit/57279809156fc089a70d9780eed958b4db3dc13c

* Quote from the Raimonds blog
http://blog.rayapps.com/2008/05/13/activerecord-oracle-enhanced-adapter/

>
Oracle adapter takes sequence values from corresponding Oracle sequences - e.g. if table name is users, then id values are taken from sequence users_seq. In original Oracle adapter sequences are created with option "START WITH 10000" which means that first id value will be 10000. I do not know why it was done so and I also did not change that in Oracle enhanced adapter. Probably one reason could be that you can use id values <10000 in your test fixtures and then also create new records in tests with automatically assigned id values and can be sure that id values will not overlap.

But this should not cause any issues for application development as id values should just be unique and it does not matter what is the starting value.
<

Since [Support reset_pk_sequence! ](https://github.com/rsim/oracle-enhanced/pull/287), there is no reason to start the primary key value from 10000 due to fixtures.

koic:
This feature was implemented by #1636.

PR #1636, opened 2018-02-03
Change `default_sequence_start_value` from 10000 to 1

yahonda:
Refer #1595 for the background of this change


koic:
This implementation is the first feature for Rails 6.0 🎉

PR #1657, opened 2018-02-17
Rails 6 requires Ruby 2.4.1+

koic:
Refer rails/rails#32034.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,5 @@
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.4
   DisabledByDefault: true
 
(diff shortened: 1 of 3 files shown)


yahonda:
Dropping JRuby 9.1.15 is fine.  By the time when Rails 6 released I expect some newer version let's say JRuby 9.2 will be released.

I was wondering why teeny versions are removed then found this pull request is following this change. 

https://github.com/rails/rails/commit/d4eb0dc89ee6b476e2e10869dc282a96f956c6c7

yahonda:
Let's restore teeny version https://github.com/rails/rails/commit/6b3acf151829ee16a946c0d77529d236f115234c

koic:
Thanks for the advice. I fixed it.

PR #1658, opened 2018-02-22
Enable `Style/ColonMethodCall` cop

yahonda:
Refer https://github.com/rails/rails/commit/9c0c90979a759a41628e0cd9d73821b0b34d03fc

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -148,3 +148,9 @@ Style/RedundantReturn:
 Style/Semicolon:
   Enabled: true
   AllowAsExpressionSeparator: true
+
+# Prefer Foo.method over Foo::method
+Style/ColonMethodCall:
+  Enabled: true
+  Exclude:
+    - 'lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb'


yahonda:
@koic I'm wondering if it is good to replace `:` with `.`  for `Java::oracle` and `Java::java` .
 
Would you review this pull request since you are familiar with RuboCop .

koic:
Yes. I also think that `Java::oracle` is customarily preferred. `Style/ColonMethodCall` cop tries to ignore Java interop code. I think this is a false positive.
https://github.com/bbatsov/rubocop/blob/v0.52.1/lib/rubocop/cop/style/colon_method_call.rb#L34-L35

It is probably better to disabling `Style/ColonMethodCall` cop  within source code as a workaround.
https://github.com/bbatsov/rubocop/blob/master/manual/configuration.md#disabling-cops-within-source-code

yahonda:
Thanks for the review.  Agreed to prefer `Java::oracle` and `Java::java`.

Also, I prefer excluding `lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb` file for this cop than adding `rubocop:disable` in the library code because I do not feel like to add rubocop code into the code itself.

koic:
>  I prefer excluding lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb file for this cop than adding rubocop:disable in the library code because I do not feel like to add rubocop code into the code itself.

Sure. You are right!

koic:
Thanks!

yahonda:
https://github.com/bbatsov/rubocop/pull/5602 addresses false positives found in this pull request. We can these lines when the newer version of RuboCop is available. https://github.com/rsim/oracle-enhanced/blob/fc19a32dc8d399212398465e9df701eb0d2d603c/.rubocop.yml#L155-L156

PR #1660, opened 2018-02-23
WIP - What happens if channel name removed

yahonda:
This pull request attempts to see which versions of RuboCop used at codeclimate if the channel removed. 

Please do not merge yet.


yahonda:
Closed. I prefer #1661

Issue #1642, opened 2018-02-05
Drop database link support


yahonda:
### Steps to reproduce

When the `nextval` branch was planned dropping database link was one of the plans.
The `nextval` branch plan itself had been canceled, I would like to propose dropping database link support from Oracle enhanced adapter in the next major version - 6.0.

### Expected behavior
Database link feature can be replaced with Rails multiple databases connection support rails/rails#28095 which is already part of Rails 5.2.

### Actual behavior
Oracle enhanced adapter needs to handle which identifiers are tables, views, and synonyms. 
However, once over the database link, all objects such as tables, views, synonyms at remote database are just identifiers.

Here are some complex code example handling database link.

https://github.com/rsim/oracle-enhanced/blob/113586d42b21825199ea81dcd402714597f3e877/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L498-L512

- Need to get `db_link` name from `describe` method

```ruby
        (owner, desc_table_name, db_link) = @connection.describe(table_name) unless owner
```

- Modify SQL string as follows

```ruby
          FROM all_triggers#{db_link}
```

### System configuration
**Rails version**: the next major version of Rails.

**Oracle enhanced adapter version**: the next major version of Oracle enhanced adapter

**Ruby version**: ruby versions supported by the next major version of Rails

**Oracle Database version**: 11.2 or higher

yahonda:
Closed per #1668

PR #1668, opened 2018-03-06
Drop db link support

yahonda:
Oracle enhanced adapter 6 drops support for db link
If users need to use database link, please consider to
- Use multiple database connections
- Create a view over the database link


yahonda:
Proposed in #1642

PR #1669, opened 2018-03-08
Drop trigger based primary key support for Rails 6

### LINKED ISSUE #1643: Drop trigger based primary key support
### Steps to reproduce
When the nextval branch was planned dropping database link was one of the plans.
The nextval branch plan itself had been canceled, I would like to propose dropping trigger based primary key support from Oracle enhanced adapter in the next major version - 6.0.

### Expected behavior
In Rails primary key value is generated by database identity data types such as MySQL `auto_increment`. Oracle enhanced adapter uses database sequence by default. Even using the sequence explicitly controlled by `prefetch_primary_key?` is implemented differently from other database adapters. 

### Actual behavior
Users can generate the primary key value based on the database triggers now.

### System configuration
**Rails version**: The next major version of Rails.

**Oracle enhanced adapter version**: The next major version of Rails.

**Ruby version**: Ruby versions supported by the next major version of Rails.

**Oracle Database version**: 11.2

yahonda:
It is planned to support identity type based primary key support
in Oracle enhanced adapter 6.
The current implementation of primary key generation is very complicated.
To make it simple, Oracle enhanced adapter 6 will drop
trigger based primary key support.

`rake db:structure:dump` still dumps triggers
if `structure_dump: db_stored_code` is configured
for those who maintain triggers by themselves.

Closes #1643


JenniferMB:
How does one transition a stable production application currently under RoR 5.2.3 and Oracle enhanced adapter 5.2.8, that allows its database to manage triggers and sequence numbers by setting sequence_name to :autogenerated, to RoR 6 and Oracle enhanced adapter 6?

yahonda:
Let me consider restoring this behavior.

matthewtusker:
Have you decided on restoring this behaviour? It has been nearly 3 years and we need this.

PR #1681, opened 2018-03-27
Remove db link codes

koic:
Follow up of #1668.

This PR removes the codes that seems to have forgotten to remove in the above.


yahonda:
Good catch. Thank you.

koic:
Thanks!

PR #1267, opened 2017-03-30
Handle `TIMESTAMP WITH TIMEZONE` separately from `TIMEZONE`

yahonda:
Fixes #1206

rails/rails#27962 moves `TypeCastingTest` from sqlite3 specific one to generic then some tests are failing because `type_cast` and `quote` method both call `quoted_date` for `Date` and `Time` class objects in abstract adapter but Oracle enhanced adapter does not, using its own implementation.

This pull request lets generic `Date` and `Type` objects handled by abstract adapter implementation.
For non-generic data types including timezone `ActiveRecord::OracleEnhanced::Type::TimestampTz` type has been created and `TimestampTz` type is handled as it used to.

Oracle enhanced adapter now relies on `Time::DATE_FORMATS` defined in ActiveSupport by this change.
Therefore changing `NLS_DATE_FORMAT` and `NLS_TIMESTAMP_FORMAT` are not working and will not be supported.

PR #1685, opened 2018-03-29
Don't allow unsupported NLS parameters

cdinger:
I ran into this issue with NLS date/timestamp formats recently. I hadn't configured any of
the [optional NLS parameters](https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L118) and knew that `nls_date_format` and `nls_timestamp_format` were unsupported, but was still running into this catastrophic error:

```
OCIError: ORA-01830: date format picture ends before converting entire input string
```

It turned out that the machine I was running on had the `NLS_TIMESTAMP_FORMAT` environment variable set and oracle_enhanced was consuming it and using the non-standard format.

This is probably not a commonly-encountered situation, but it was very confusing and time-consuming to track down for me. This PR I think makes this restriction clearer if you choose to accept it.

Since this is an ActiveRecord adapter, it seems OK to force these parameters (`NLS_TIMESTAMP_FORMAT` and `NLS_DATE_FORMAT`) to the format Rails expects; I can't think of a use case for changing these values if it fundamentally breaks ActiveRecord.

Thanks for considering!

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
@@ -161,6 +161,10 @@ module ActiveRecord
             end
           end
 
+          OracleEnhancedAdapter::FIXED_NLS_PARAMETERS.each do |key, value|
+            exec "alter session set #{key} = '#{value}'"
+          end
+
           self.autocommit = true
 
           schema = config[:schema] && config[:schema].to_s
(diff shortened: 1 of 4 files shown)


yahonda:

Would you address CI failure with jruby-head https://travis-ci.org/rsim/oracle-enhanced/builds/360041148?utm_source=github_status&utm_medium=notification so that I can review.

Thank you.

cdinger:
@yahonda sorry about that—I neglected to run the specs in jruby. It's green now. Thanks for looking.

yahonda:
Thanks for the update. Let me review it.
Note: background for this pull request #1267 for my reminder to review.

yahonda:
Merged. 

I was wondering if we need to show some message who configures and expects non-default values for `nls_date_format` and `nls_timestamp_format` values but it would be too much. 

Thanks again for the pull request.

PR #1694, opened 2018-04-10
Update README for Rails 5.2 [skip ci]

koic:
Follow up of #1637.

@yahonda Thank you for releasing Oracle enhanced adapter 5.2.0 🎉


yahonda:
Good catch and thank you.

Issue #1505, opened 2017-09-28
Exclude `materialized_views` from `tables`


yahonda:
### Steps to reproduce

`tables` method returns both tables and materialized views because Oracle database `all_tables` dictionary returns materialized views also.

I'd like to propose to exclude materialized views from `tables` method and add `materialized_views` to `data_source`. 

* `tables`, `views` and `materialized_views` methods

```ruby

      def tables #:nodoc:
        select_values(<<-SQL, "SCHEMA")
          SELECT DECODE(table_name, UPPER(table_name), LOWER(table_name), table_name)
          FROM all_tables WHERE owner = SYS_CONTEXT('userenv', 'current_schema') AND secondary = 'N'
        SQL
      end
```

```ruby
      def views # :nodoc:
        select_values("SELECT LOWER(view_name) FROM all_views WHERE owner = SYS_CONTEXT('userenv', 'current_schema')")
      end
```

```ruby
      def materialized_views #:nodoc:
        select_values("SELECT LOWER(mview_name) FROM all_mviews WHERE owner = SYS_CONTEXT('userenv', 'current_schema')")
      end

```

Example using sql statements used in each methods:

```sql
SQL>  CREATE TABLE "TEST_POSTS" ("POST_ID" NUMBER(38) NOT NULL PRIMARY KEY, "TITLE" VARCHAR2(255), "CREATED_AT" TIMESTAMP NULL, "UPDATED_AT" TIMESTAMP NULL)
  2  ;

Table created.

SQL> CREATE MATERIALIZED VIEW test_posts_mv AS SELECT * FROM test_posts;

Materialized view created.
```

* `tables` returns both `test_posts` table and `test_posts_mv` materialized view

```
SQL> SELECT DECODE(table_name, UPPER(table_name), LOWER(table_name), table_name)
  2  FROM all_tables WHERE owner = SYS_CONTEXT('userenv', 'current_schema') AND secondary = 'N';

DECODE(TABLE_NAME,UPPER(TABLE_NAME),LOWER(TABLE_NAME),TABLE_NAME)
--------------------------------------------------------------------------------
test_posts
test_posts_mv

SQL>
```

* `views` returns no rows, materialized views are not views.

```sql
SQL> SELECT LOWER(view_name) FROM all_views WHERE owner = SYS_CONTEXT('userenv', 'current_schema');

no rows selected
```

* `materialized_views` returns `test_posts_mv` materialized views

```sql
[...]

PR #1708, opened 2018-04-17
`OracleEnhanced::SchemaStatements#tables` excludes materialized views

yahonda:
If users want to see materialized views, call `OracleEnhanced::SchemaStatements#materialized_views`

Addresses #1505


KyGuyMayor:
@yahonda Can you give an example of how to use materialized views after this change?

PR #1710, opened 2018-04-24
CI with JRuby 9.1.17.0

yahonda:
http://jruby.org/2018/04/23/jruby-9-1-17-0.html


yahonda:
FYI https://github.com/rvm/rvm/pull/4366

PR #1715, opened 2018-04-27
Remove `rails/arel` from bug report templates [skip ci]

koic:
Arel is now bundled in the Active Record gem, and maintained in the rails/rails repository.

Related PR is #1711.


yahonda:
Good catch. Well, we see two issues reported by codeclimate. I do not think we need to "fix" these failures and also I do not prefer getting this kind of failures. I have merged #1716 to expect these failures disappeared. Would you rebase the current master branch and push -f the same commit again to run the check again.

koic:
Thanks for helping. I rebased it and Code Climate was green.

PR #1724, opened 2018-05-22
Enable `Lint/StringConversionInInterpolation` cop

koic:
Follow up of https://github.com/rails/rails/commit/a77447f.


yahonda:
Thank you.

PR #1744, opened 2018-08-01
Introduce column cache per connection

yahonda:
rsim/oracle-enhanced#1490 removed Oracle enhanced adapter own column cache feature
to use Rails `db:schema:cache:dump`.  However, it caused slower performance
about metadata search, such as column information.

This commit restores Oracle enhanced adapter own column cache feature.
It used to be class level caching, it is connection level caching now
since it is much easier and cleaner than class level ones.

Related to rsim/oracle-enhanced#1720

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -157,6 +157,14 @@ module ActiveRecord
           all_schema_indexes.select { |i| i.table == table_name }
         end
 
+        def columns(table_name)
+          table_name = table_name.to_s
+          if @columns_cache[table_name]
+            @columns_cache[table_name]
+          else
+            @columns_cache[table_name] = super(table_name)
+          end
+        end
         # Additional options for +create_table+ method in migration files.
         #
         # You can specify individual starting value in table creation migration file, e.g.:
@@ -253,6 +261,8 @@ module ActiveRecord
           execute "DROP SEQUENCE #{quote_table_name(seq_name)}" rescue nil
         rescue ActiveRecord::StatementInvalid => e
           raise e unless options[:if_exists]
+        ensure
+          clear_table_columns_cache(table_name)
         end
 
         def insert_versions_sql(versions) # :nodoc:
@@ -403,6 +413,8 @@ module ActiveRecord
           execute add_column_sql
           create_sequence_and_trigger(table_name, options) if type && type.to_sym == :primary_key
           change_column_comment(table_name, column_name, options[:comment]) if options.key?(:comment)
+        ensure
+          clear_table_columns_cache(table_name)
         end
 
         def aliased_types(name, fallback)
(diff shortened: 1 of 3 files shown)


yahonda:
@koic I'd like you to review this pull request since this implementation just works, but does not look good.

https://github.com/yahonda/oracle-enhanced/blob/4a1ed1116b3027896c113531b9ce61764efdf891/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb#L159-L170

```ruby
        def columns(table_name)
          table_name = table_name.to_s
          if @columns_cache[table_name]
            @columns_cache[table_name]
          else
            @columns_cache[table_name] =
              column_definitions(table_name).map do |field|
              new_column_from_field(table_name, field)
            end
          end
          @columns_cache[table_name]
        end
```

Reasons:

* Copy and paste super method implementation https://github.com/rails/rails/blob/fee7fba8934ee6005b7e8eefc4baf7db48f03934/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L110-L116

* `@columns_cache[table_name]` appears many times.

Any feedbacks are welcomed.

koic:
Sure thing. I will look it next week. Please wait a moment.

lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
```diff
@@ -156,6 +156,18 @@ def indexes(table_name) #:nodoc:
...
+          if @columns_cache[table_name]
+            @columns_cache[table_name]
+          else
+            @columns_cache[table_name] =
+              column_definitions(table_name).map do |field|
+              new_column_from_field(table_name, field)
+            end
+          end
+          @columns_cache[table_name]
+        end
```
koic: How about using `super` to transfer to `AR::ConnectionAdapters::SchemaStatements#columns(table_name)` as follows?

```diff
         def columns(table_name)
           table_name = table_name.to_s
           if @columns_cache[table_name]
             @columns_cache[table_name]
           else
+            @columns_cache[table_name] = super(table_name)
-            @columns_cache[table_name] =
-              column_definitions(table_name).map do |field|
-              new_column_from_field(table_name, field)
-            end
           end
-          @columns_cache[table_name]
         end
```

There will be no duplication of implementation with super class.

koic:
> Copy and paste super method implementation

I wrote a comment on https://github.com/rsim/oracle-enhanced/pull/1744/files#r208188648.

> @columns_cache[table_name] appears many times.

On the other hand. I don't have an idea to this dup codes...

yahonda:
@koic Thanks for the review. Making use of `super` looks good. I have made a change as you suggested and squashed.

lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
```diff
@@ -157,6 +157,15 @@ def indexes(table_name) #:nodoc:
...
         end
 
+        def columns(table_name)
+          table_name = table_name.to_s
+          if @columns_cache[table_name]
+            @columns_cache[table_name]
+          else
+            @columns_cache[table_name] = super(table_name)
+          end
+          @columns_cache[table_name]
```
koic: I think that this explicit return value line is unnecessary. The return value is determined by L163 or L165.

yahonda: Right. Updated.

koic:
Thanks for fixing pragmatic issue!

Issue #1680, opened 2018-03-22
Unexpected behavior when using bind params for CHAR column


agungl:
### Steps to reproduce
Create a table with a column of type CHAR(10) as primary key and some other columns.
Construct an ActiveRecord instance related to this table and _id = "ABC"_, then save it to the DB.
Update another attribute and _call_ save to update the record in the DB.

### Expected behavior
The existing record in the DB should get updated and reflect the change of the other attribute.

### Actual behavior
The record is not updated. The problem might be, that the parameter binding for the CHAR column uses a binding for type VARCHAR2. CHAR would be correct which blank pads the "ABC" (length = 3) to "ABC       " (length = 10). Only then the existing record would be found and updated.
Find more details about it at: https://asktom.oracle.com/pls/asktom/f?p=100:11:0::::P11_QUESTION_ID:656026400346262738
The current behavior makes working with legacy schemas containing many CHAR columns difficult.

### System configuration
**Rails version**: 5.1.5

**Oracle enhanced adapter version**: 1.8.0

**Ruby version**: jruby-9.1.13.0

**Oracle Database version**: 12.1.0.2

yahonda:
Hi, As I have asked #1679 would you create an executable test case to reproduce it? 

Thanks.

agungl:
Here is a testcase:

[testcase1680.txt](https://github.com/rsim/oracle-enhanced/files/1840613/testcase1680.txt)

yahonda:
Thanks for providing the test case. As you already aware this error only reproduces if you add `CHAR` column manually. Oracle enhanced adapter maps `string` type to `VARCHAR2` sql type. 

https://github.com/rsim/oracle-enhanced/blob/f843447cdd626c295aa7d00dfd09000c7b9a3abd/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L333

I understand there are some cases who want to use `CHAR` but I would like to say that users who need non-default configuration will need to investigate and handle these cases by themselves. You may dig in  `bind_params` method . 

https://github.com/rsim/oracle-enhanced/blob/f843447cdd626c295aa7d00dfd09000c7b9a3abd/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb#L311-L344

Thank you.

agungl:
[...]

PR #1760, opened 2018-08-20
fix Handling with char-column problem

mimiyanwk:
the problem with processing with char column is fixed


yahonda:
Thanks for opening a pull request. Would you add some test cases? As far as I remember there was a test case explained in #1680. Thank you.

yahonda:
Thanks for updating specs. Would you address RuboCop offenses reported at https://codeclimate.com/github/rsim/oracle-enhanced/pull/1760 ?

mimiyanwk:
I have added a test case explained in #1680 but have a question:
Char was mapped to String in ruby-oci8.
How can I check if a String type of result set is from Char in oci_connection.rb?

mimiyanwk:
The test case explained in #1680 is now completely implemented.

yahonda:
Thanks. Would you squash commits into one?

mimiyanwk:
Squashing commits into one was finisched.

yahonda:
Thank you.

PR #1765, opened 2018-09-06
table with quoted name (case sensitive) and synonym

g-portugues:
I've a database that have many tables with quoted name (case sensitive). It's a problem because when I use a model class to make reference to the table and I use `self.table_name = 'MYSCHEMA."mytable"'` I get this error: `ActiveRecord::ConnectionAdapters::OracleEnhanced::ConnectionException ("DESC SPDNET."config_tmp"" failed; does it exist?)`. This table have a **synonym**, but, if I use this synonym in `table_name` of the model class, then the gem stay in loop in the method `OracleEnhanced::Connection.describe` and generate this error: `SystemStackError (stack level too deep)`.
Because this, I changed `OracleEnhanced::Connection.describe` to verify if table_name of synonym is case sensitive or not and replace **"** when search a database object (table, view or synonym) in Oracle's system tables.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
@@ -24,11 +24,8 @@ module ActiveRecord
           def describe(name)
             name = name.to_s
             if name.include?("@")
-              name, db_link = name.split("@")
-              default_owner = _select_value("SELECT username FROM all_db_links WHERE db_link = '#{db_link.upcase}'")
-              db_link = "@#{db_link}"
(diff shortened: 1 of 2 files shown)


yahonda:
Would you add some specs to make sure this code change works as expected? 

Thank you.

g-portugues:
> Thanks for opening a pull request.
> Would you add some specs to make sure this code change works as expected?
> 
> Thank you.

Ok, I did this today... sorry for the delay.

yahonda:
Thanks for adding a test. Let me have some time to review it.

PR #1773, opened 2018-10-17
Add a check to supported Oracle database version

koic:
Follow up of rails/rails#34232.

This PR adds a check to supported Oracle database version.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -644,6 +644,13 @@ module ActiveRecord
       alias index_name_length max_identifier_length
 
       private
+        def check_version
+          database_version = @connection.database_version.join(".").to_f
+
+          if database_version < 10
+            raise "Your version of Oracle (#{database_version}) is too old. Active Record Oracle enhanced adapter supports Oracle >= 10g."
+          end
+        end
 
         def initialize_type_map(m = type_map)
           super


lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
```diff
@@ -644,6 +644,13 @@ def max_identifier_length
       alias index_name_length max_identifier_length
 
       private
+        def check_version
+          database_version = @connection.database_version.join(".").to_f
+
+          if database_version < 10
+            raise "Your version of Oracle (#{database_version}) is too old. Active Record supports Oracle >= 10g."
```
yahonda: Active Record itself does not "support" Oracle database, Active Record Oracle enhanced adapter does. Then, "Active Record Oracle enhanced adapter support Oracle" may be a good idea.

koic: Sure! Oracle enhanced adapter is a 3rd party adapter. I applied your suggestion to this message.

koic:
Thanks for the merging this PR.
I specified Oracle 10g according to the supported minimum Oracle database version written in the following README.md.

> Oracle enhanced adapter can be used from Ruby on Rails versions between 2.3.x and 5.1 and it is working with Oracle database versions from 10g to 12c.

https://github.com/rsim/oracle-enhanced/blob/v5.2.3/README.md

I think that it maybe good to consider the supported minimum Oracle database version for Rails 6 another issue. Thank you!

yahonda:
Agreed. I'll send a proposal email bumping the minimum version of Oracle database to somewhat newer to https://groups.google.com/forum/#!forum/oracle-enhanced

> I think that it maybe good to consider the supported minimum Oracle database version for Rails 6 another issue. Thank you!

yahonda:
FYI: https://groups.google.com/forum/#!topic/oracle-enhanced/-G38mUJopFw

PR #1789, opened 2018-11-23
Adds xml type support

abhikur:
This adds support for xml column to store a ruby hash. For example it will store `{key: 'value'}` to oracle xml type column as `<key>value</key>` and vice-versa.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ group :development do
   gem "rspec"
   gem "rdoc"
   gem "rake"
-  gem 'xmlhasher'
+  gem "xmlhasher"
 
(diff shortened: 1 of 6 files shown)


yahonda:
Thanks for opening a pull request. CI with JRuby does not pass.

https://travis-ci.org/rsim/oracle-enhanced/jobs/458706476

Oracle enhanced adapter supports both CRuby and JRuby. It looks like `escape_utils` gem which is required by `xmlhasher`, only supports CRuby. Do you have any idea some other gems can run on CRuby and JRuby?

abhikur:
Hi, yes you are correct. I was looking for the alternative of XMLHasher.
Will resolve the problem and then push the changes.

PR #1795, opened 2018-12-05
CI against JRuby 9.2.4.1

koic:
JRuby 9.2.4.1 has been released.
https://www.jruby.org/2018/11/28/jruby-9-2-4-1


yahonda:
Thanks.

PR #1802, opened 2018-12-21
Enable `Style/RedundantBegin` cop

koic:
Follow up of rails/rails#34764.


koic:
It seems that the jruby matrix failed for reasons not related to this PR change.
https://travis-ci.org/rsim/oracle-enhanced/jobs/470981347#L804-L810

I think that there is a problem with the combination of I18n gem and JRuby. This requires investigation separately.

koic:
@yahonda I rebased this PR and CI succeeded. Thanks for investigating and applying the workaround patch.

Issue #1737, opened 2018-07-29
ORA-12899: value too large for column "FIELD_WITH_LEADING_SPACE" (actual: 1022, maximum: 1020)


yahonda:
### Steps to reproduce
```ruby
$ git checkout release52
$ bundle exec rspec ./spec/active_record/connection_adapters/oracle_enhanced/schema_dumper_spec.rb:379
```
### Expected behavior
It should pass.

### Actual behavior
```ruby
$ bundle exec rspec ./spec/active_record/connection_adapters/oracle_enhanced/schema_dumper_spec.rb:379
Coverage may be inaccurate; set the "--debug" command line option, or do JRUBY_OPTS="--debug" or set the "debug.fullTrace=true" option in your .jrubyrc
/home/yahonda/.rbenv/versions/jruby-9.2.1.0-dev/lib/ruby/gems/shared/bundler/gems/simplecov-03a9f7fde44a/lib/simplecov.rb:53: warning: tracing (e.g. set_trace_func) will not capture all events without --debug flag
==> Loading config from ENV or use default
==> Running specs with JRuby version 9.2.1.0-SNAPSHOT
/home/yahonda/.rbenv/versions/jruby-9.2.1.0-dev/lib/ruby/gems/shared/bundler/gems/ruby-plsql-e0b3e01204a8/lib/plsql/jdbc_connection.rb:225: warning: constant ::Fixnum is deprecated
/home/yahonda/.rbenv/versions/jruby-9.2.1.0-dev/lib/ruby/gems/shared/bundler/gems/ruby-plsql-e0b3e01204a8/lib/plsql/jdbc_connection.rb:226: warning: constant ::Bignum is deprecated
==> Effective ActiveRecord version 5.2.0
Run options: include {:locations=>{"./spec/active_record/connection_adapters/oracle_enhanced/schema_dumper_spec.rb"=>[379]}}
/home/yahonda/git/oracle-enhanced/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb:241: warning: constant ::NativeException is deprecated
/home/yahonda/git/oracle-enhanced/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb:460: warning: constant ::NativeException is deprecated
F
An error occurred in an `after(:context)` hook.
Failure/Error: s.execute(sql)

ActiveRecord::StatementInvalid:
  NativeException: java.sql.SQLSyntaxErrorException: ORA-00942: table or view does not exist
[...]

PR #1807, opened 2018-12-26
CI against Ruby 2.6 and bump other Ruby versions

yahonda:
Kind of backporting #1806 to release52 and revert #1740 since #1737 had been addressed.


yahonda:
So far I have no plan to backport this pull request to older versions of Oracle enhanced adapter.

PR #1815, opened 2019-01-18
Remove deprecated `#insert_fixtures` method for Rails 6.0

koic:
Follow up of https://github.com/rails/rails/commit/400ba786e1d154448235f5f90183e48a1043eece.


yahonda:
Thanks for opening a pull request. Let me confirm if we can remove this method from Oracle enhanced adapter there may be Oracle enhanced adapters fixture implementation differences.

yahonda:
Confirmed fixtures are working fine with this pull request.

koic:
Thanks for the confirmation

PR #1821, opened 2019-01-31
CI against Ruby 2.6.1

yahonda:
https://www.ruby-lang.org/en/news/2019/01/30/ruby-2-6-1-released/

It should be merged when https://github.com/rvm/rvm/pull/4597 or equivalent commit is ready.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
 
 language: ruby
 rvm:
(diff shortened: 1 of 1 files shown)


yahonda:
Looks like Ruby 2.6.1 is installed without rvm/rvm#4597 .

PR #1839, opened 2019-03-05
fix mapping of decimal type to map to `NUMBER` instead of `DECIMAL`

punkisdead:
This patch to the `OracleEnhancedAdapter` is necessary because of the way that the gem generates the SQL to create
tables with `decimal` columns. Without the patch, if you were to create a migration that defined a table with a `decimal`
column without specifiying a `precision` and `scale` such as:

    class AddExampleTable < ActiveRecord::Migration[5.2]
      def change
        create_table :example do |t|
          t.string :name
          t.decimal :value
        end
      end
    end

The OracleEnhancedAdapter would generate the following SQL:

    CREATE TABLE example (
            id NUMBER(5) PRIMARY KEY,
            name VARCHAR2(15) NOT NULL,
            value DECIMAL)

This is problematic because Oracle then translates the `DECIMAL` datatype to `NUMBER(18,0)` effectively making this an `INTEGER`.


punkisdead:
Older versions of this gem used to correctly map `t.decimal` to the Oracle type `NUMBER()`. Not sure when it changed, but we just upgraded our app this past week and noticed this strange behavior.

Basically if we defined a column as being `t.decimal` with no `precision` or `scale`, by the time the migration was run and the `schema.rb` was re-generated, it would then show up as a `t.integer` in the `schema.rb`.

PR #1846, opened 2019-03-11
Add spec for custom sequence name

yahonda:
[MigrationTest#test_create_table_with_custom_sequence_name ](https://github.com/rails/rails/blob/50e01ec01a7658ba4542bb5f9b68223e98ce6180/activerecord/test/cases/migration_test.rb#L570-L602) is only executed for `OracleAdapter`, it would be nice to run this test at Oracle enhanced adapter.

As Oracle enhanced adapter with Oracle Database 12c supports longer identifier,
this pull request does not test 29 byte length table name and 33-byte length sequence name.

Will open a pull request at Rails to remove `MigrationTest#test_create_table_with_custom_sequence_name`

### DIFF
--- a/spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
+++ b/spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
@@ -556,4 +556,26 @@ describe "OracleEnhancedAdapter" do
       Thread.report_on_exception = @original_report_on_exception
     end
   end
+
+  describe "Sequence" do
+    before(:all) do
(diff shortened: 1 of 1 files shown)


yahonda:
@koic Would you take a look at this pull request and give some feedbacks?

yahonda:
@koic I should have explained why I requested a review. Let me discuss this afternoon.

koic:
@yahonda First of all, I think it's good to start with the existing minimal rails/rails test case import. So this looks good to me.

yahonda:
Thanks for the review.

PR #1856, opened 2019-03-27
Use RuboCop 0.66

yahonda:
Looks like `rubocop-0-66` channel is available at Code Climate.

PR #1857, opened 2019-03-27
Bump RuboCop to 0.66.0

koic:
Follow up https://github.com/rails/rails/pull/35761.


koic:
Dup #1856.

yahonda:
Oh, I was not aware that you already have opened https://github.com/rails/rails/pull/35761 and this one. Thanks for the review at #1856.

koic:
Thank you for very fast catch!

PR #1858, opened 2019-04-01
CI against Ruby 2.4.6

yahonda:
https://www.ruby-lang.org/en/news/2019/04/01/ruby-2-4-6-released/


yahonda:
Refer https://github.com/rvm/rvm/pull/4657

PR #1860, opened 2019-04-09
Except `table_name` from column objects

koic:
Follow up https://github.com/rails/rails/pull/35890.

This PR fixes the following error.

```consle
% cd path/to/oracle-enhanced
% bundle exec rake

(snip)

Failures:

  1) OracleEnhancedAdapter schema dump virtual columns should dump
  correctly
     Failure/Error: expect(output).to match(/t\.virtual
  "full_name",(\s*)type: :string,(\s*)limit: 512,(\s*)as:
  "\\"FIRST_NAME\\"\|\|', '\|\|\\"LAST_NAME\\""/)

       expected "# This file is auto-generated from the current state of
       the database. Instead\n# of editing this fil...e_name' for
       #<ActiveRecord::ConnectionAdapters::OracleEnhanced::Column:0x0000000003a21bd0>\n\nend\n"
       to match /t\.virtual "full_name",(\s*)type: :string,(\s*)limit:
       512,(\s*)as: "\\"FIRST_NAME\\"\|\|', '\|\|\\"LAST_NAME\\""/
       Diff:
       @@ -1,2 +1,19 @@
       -/t\.virtual "full_name",(\s*)type: :string,(\s*)limit:
       512,(\s*)as: "\\"FIRST_NAME\\"\|\|', '\|\|\\"LAST_NAME\\""/
       +# This file is auto-generated from the current state of the
       database. Instead
       +# of editing this file, please use the migrations feature of
       Active Record to
       +# incrementally modify your database, and then regenerate this
       schema definition.
       +#
       +# This file is the source Rails uses to define your schema when
       running `rails
       +# db:schema:load`. When creating a new database, `rails
       db:schema:load` tends to
       +# be faster and is potentially less error prone than running all
       of your
       +# migrations from scratch. Old migrations may fail to apply
       correctly if those
       +# migrations use external dependencies or application code.
       +#
       +# It's strongly recommended that you check this file into your
       version control system.
       +
       +ActiveRecord::Schema.define(version: 0) do
       +
       +# Could not dump table "test_names" because of following
       NoMethodError
       +#   undefined method `table_name' for
       #<ActiveRecord::ConnectionAdapters::OracleEnhanced::Column:0x0000000003a21bd0>
       +
       +end
```

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/column.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/column.rb
@@ -6,8 +6,8 @@ module ActiveRecord
       class Column < ActiveRecord::ConnectionAdapters::Column
         delegate :virtual, to: :sql_type_metadata, allow_nil: true
 
-        def initialize(name, default, sql_type_metadata = nil, null = true, table_name = nil, comment = nil) #:nodoc:
-          super(name, default, sql_type_metadata, null, table_name, comment: comment)
+        def initialize(name, default, sql_type_metadata = nil, null = true, comment: nil) #:nodoc:
+          super(name, default, sql_type_metadata, null, comment: comment)
         end
 
         def virtual?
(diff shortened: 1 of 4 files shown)


yahonda:
Thanks for the fix.

PR #1861, opened 2019-04-10
Remove unused `sequence_name` in `sql_for_insert`

koic:
Follow up https://github.com/rails/rails/commit/f21bc468c35c295351309aada83ccbe44738e653.

This PR fixes the following error.

```console
% bundle exec rake

(snip)

  265) OracleEnhancedAdapter timestamp with timezone support / TIMESTAMP
  WITH TIME ZONE values from ActiveRecord model should return Time value
  from TIMESTAMP columns
       Failure/Error: super

       ArgumentError:
         wrong number of arguments (given 4, expected 3)
       #
         /home/vagrant/.rvm/gems/ruby-2.6.2/bundler/gems/rails-7cb3e8b8efb1/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb:490:in
         `sql_for_insert'
       #
         ./lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb:84:in
         `sql_for_insert'
       #
         ./lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb:94:in
         `exec_insert'
```

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb
@@ -76,7 +76,7 @@ module ActiveRecord
 
         # New method in ActiveRecord 3.1
         # Will add RETURNING clause in case of trigger generated primary keys
-        def sql_for_insert(sql, pk, sequence_name, binds)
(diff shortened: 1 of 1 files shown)


yahonda:
Thanks!

PR #1869, opened 2019-04-16
Use squiggly heredoc to strip odd indentation in the executed SQL

koic:
Follow up https://github.com/rails/rails/commit/1b68ead8f8cb0d7a1771ab82ab20bd79857b40ee.


lib/active_record/connection_adapters/oracle_enhanced/connection.rb
```diff
@@ -34,27 +34,27 @@ def describe(name)
...
-          FROM all_synonyms
-          WHERE owner = '#{table_owner}'
-            AND synonym_name = '#{table_name}'
-          UNION ALL
-          SELECT table_owner, table_name, 'SYNONYM' name_type
-          FROM all_synonyms
-          WHERE owner = 'PUBLIC'
-            AND synonym_name = '#{real_name}'
-        SQL
+            sql = <<~SQL.strip.gsub(/\s+/, " ")
```
koic: `strip.gsub(/\s+/, " ")` is left for non-breaking query logging 📝

yahonda:
@koic Please go ahead to merge this pull request. I'm curious if it is a good idea to implement a Cop to enforce squiggly heredoc.

koic:
Hmm, I think the `Layout / IndentHeredoc` cop is that, but it seems a different.
https://rubocop.readthedocs.io/en/stable/cops_layout/#layoutindentheredoc

Let's talk about that on RubyKaigi 2019 if we have a chance.

PR #1874, opened 2019-05-16
Add TCP keepalive and TCP keepalive time configurations

koic:
## Summary

Related with #1489.

In Active Record Oracle enhanced adapter 5.2, an Oracle error occurs when TCP keepalive time exceeds 600 seconds.

The following ORA-03113 error occurred every 10 minutes when I tried to upgrade from Rails 5.1.6.1 (activerecord-oracle_enhanced-adapter 1.8.2) to Rails 5.2.2 (activerecord-oracle_enhanced-adapter 5.2.5).

```console
stmt.c:243:in oci8lib_240.so: OCIError: ORA-03113: 通信チャネルでend-of-fileが検出されました (ActiveRecord::StatementInvalid)プロセスID: 24578
```

I think that TCP Keepalive parameters depend on an application purpose.
Therefore this PR allows users to change TCP keepalive parameters.

The default behavior has not changed from Active Record Oracle enhanced adapter 5.2 so that breaking changes don't occur.

### Other Information

TCP keepalive parameter names conform to ruby-oci8.
http://www.rubydoc.info/github/kubo/ruby-oci8/file/docs/hanging-after-inactivity.md

PR #1878, opened 2019-05-16
Add TCP keepalive and TCP keepalive time configurations

yahonda:
This pull request backports #1874 to release52 branch.


yahonda:
Note: Although, I usually do not backport new features into stable branches like release52.  I think this is a kind of regression introduced by #1489 and this backport should not affect regressions and/or incompatibilities to the stable branch.

PR #1890, opened 2019-06-13
Enable `Layout/EmptyLinesAroundAccessModifier` cop

koic:
Follow up https://github.com/rails/rails/pull/36472.

This PR enables `Layout/EmptyLinesAroundAccessModifier` cop with `EnforcedStyle: only_before`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,6 +30,10 @@ Layout/ElseAlignment:
 Layout/EmptyLineAfterMagicComment:
   Enabled: true
 
+Layout/EmptyLinesAroundAccessModifier:
+  Enabled: true
+  EnforcedStyle: only_before
+
 Layout/EmptyLinesAroundBlockBody:
   Enabled: true
 
--- a/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
@@ -19,7 +19,6 @@ module ActiveRecord
         attr_reader :raw_connection
 
         private
-
           # Used always by JDBC connection as well by OCI connection when describing tables over database link
           def describe(name)
             name = name.to_s
--- a/lib/active_record/connection_adapters/oracle_enhanced/context_index.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/context_index.rb
@@ -142,7 +142,6 @@ module ActiveRecord
         end
 
         private
-
           def create_datastore_procedure(table_name, procedure_name, column_names, options)
             quoted_table_name = quote_table_name(table_name)
             select_queries, column_names = column_names.partition { |c| c.to_s =~ /^\s*SELECT\s+/i }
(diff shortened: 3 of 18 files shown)


yahonda:
Restarted `continuous-integration/travis-ci/pr` without investigation yet. Will see how it works.

yahonda:
These failures can be addressed by reverting https://github.com/rails/rails/commit/c5ecc338e863e771adcbcd87d5c23ae5e7ca7894 but I do not know why.

kamipo:
Ensure "schema_migrations" table existance before each spec will fix the failures.

Since https://github.com/rails/rails/commit/c5ecc338e863e771adcbcd87d5c23ae5e7ca7894, `SchemaMigration.table_exists?` respect schema cache like as others (i.e. each `table_exists?` call doesn't always hit the database), so if "schema_migrations" table will be dropped before running spec in transactional tests for some reason, it may cause that create "schema_migrations" table -> hit schema cache -> rolled back "schema_migrations" table creation (but still schema cache exists).

yahonda:
Thanks for the information. 

I was investigating and getting the same conclusion.  While `AR_INTERNAL_METADATA` table exists at the database, `connection.schema_cache.data_source_exists?(table_name)` returns `false` then these spec attempt to create `AR_INTERNAL_METADATA` table incorrectly.

```ruby
From: /home/yahonda/git/rails/activerecord/lib/active_record/model_schema.rb @ line 342 ActiveRecord::ModelSchema::ClassMethods#table_exists?:

    339: def table_exists?
    340:   require 'pry'
    341:   require 'pry-nav'
 => 342:   binding.pry unless connection.schema_cache.data_source_exists?(table_name)
    343:   connection.schema_cache.data_source_exists?(table_name)
    344: end

[1] pry(ActiveRecord::InternalMetadata)> table_name
=> "ar_internal_metadata"
[2] pry(ActiveRecord::InternalMetadata)> connection.schema_cache.data_source_exists?(table_name)
=> false
[3] pry(ActiveRecord::InternalMetadata)> connection.data_source_exists?(table_name)
=> true
```

```sql
SQL> select table_name from user_tables;

TABLE_NAME
--------------------------------------------------------------------------------
POSTS
AR_INTERNAL_METADATA
```

@kamipo 
I do not think these errors occur at user applications, however as a 3rd party Active Record connection adapter maintainer who needs to run unit tests frequently creating and dropping database tables, `connection.table_exists?(table_name)` and `connection.schema_cache.data_source_exists?(table_name)` are not duplicate. Since the first one returns the result based on the database and the later one returns the schema cache. 

It would be appreciated if you consider restoring the original behavior or adding some deprecation cycle for this change.

kamipo:
If schema cache is incorrect with the database, it already is something wrong, we already have clearing schema cache (`reset_column_information`) in much places, when changing schema after schema cache hit.
Since I've found to be able to test easily that fact, I've added clearing schema cache in `create_table`, `drop_table`, `rename_table` first.

yahonda:
Thanks for the useful information. Applying https://github.com/rails/rails/commit/5cab344494c340ea82a35b46efa06b94f0b7730b to Oracle enhanced adapter https://github.com/yahonda/oracle-enhanced/commit/b1e04aaf34d4e025764b9d31a3d1e7ce7ba03de3 resolves these errors.

yahonda:
@koic #1891 has been merged, then would you rebase this pull request based on the current master branch?

koic:
Thank you everyone!

PR #1891, opened 2019-06-13
Clear schema cache when a table is created/dropped/renamed

kamipo:
Refer https://github.com/rails/rails/commit/5cab344494c340ea82a35b46efa06b94f0b7730b

It is required for passing spec with master in rails/rails after
https://github.com/rails/rails/commit/c5ecc338e863e771adcbcd87d5c23ae5e7ca7894.


yahonda:
Thanks!

PR #1894, opened 2019-06-19
Clear schema cache when a table is created/dropped/renamed

yahonda:
Kind of backporting #1891 to release60


yahonda:
I was misunderstanding the error. #1895 should make Ci green.

PR #1900, opened 2019-07-09
Raise `NoDatabaseError` when db does not exist

koic:
Follow up https://github.com/rails/rails/pull/13469 and https://github.com/rails/rails/pull/13427.

This PR raises `NoDatabaseError` when db does not exist.

`OCIError` was raised when database does not exist.
This change uses `NoDatabaseError` as other core adapters.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -73,6 +73,12 @@ module ActiveRecord
         ConnectionAdapters::OracleEnhancedAdapter.new(
           ConnectionAdapters::OracleEnhanced::Connection.create(config), logger, config)
       end
+    rescue OCIError => error
+      if /ORA-12154/.match?(error.message)
+        raise ActiveRecord::NoDatabaseError
+      else
+        raise
+      end
     end
   end
 
--- a/spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
+++ b/spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
@@ -244,6 +244,14 @@ describe "OracleEnhancedAdapter" do
     end
   end
 
+  describe "oracle_enhanced_connection" do
+    it "should bad connection" do
+      expect {
+        ActiveRecord::Base.oracle_enhanced_connection(adapter: "oracle_enhanced", database: "should_not_exist-cinco-dog-db")
+      }.to raise_error(ActiveRecord::NoDatabaseError)
+    end
+  end
+
   describe "explain" do
     before(:all) do
       @conn = ActiveRecord::Base.connection


yahonda:

From my point of view `ORA-12154` does not always mean database does not exist.
It just means TNS alias cannot be resolved, then it could reproduce database just exists but not started up and something like that. 

https://docs.oracle.com/en/database/oracle/oracle-database/19/errmg/ORA-12150.html#GUID-FF3C136B-78D5-4573-BD45-7D9CAC2A1C0F

```
ORA-12154: TNS:could not resolve the connect identifier specified
Cause: A connection to a database or other service was requested using a connect identifier, and the connect identifier specified could not be resolved into a connect descriptor using one of the naming methods configured. For example, if the type of connect identifier used was a net service name then the net service name could not be found in a naming method repository, or the repository could not be located or reached.
```

lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
```diff
@@ -73,6 +73,12 @@ def oracle_enhanced_connection(config) #:nodoc:
         ConnectionAdapters::OracleEnhancedAdapter.new(
           ConnectionAdapters::OracleEnhanced::Connection.create(config), logger, config)
       end
+    rescue OCIError => error
```
koic: Oops. It should also support JDBC.

> expected ActiveRecord::NoDatabaseError, got #<NameError: uninitialized constant ActiveRecord::ConnectionHandling::OCIError> with backtrace:

https://travis-ci.org/rsim/oracle-enhanced/jobs/556237777#L651

yahonda:
Let me have some more look at which error is appropriate to represent the database does not exist.

yahonda:
### Summary:

From my point of view there is no good Oracle error message which represents `NoDatabaseError` exception. ORA-01017 is a candidate one but it could make users misunderstanding the cause of this `NoDatabaseError` exception, then I prefer not to implements exception yet.

### Backgrounds:

As far as I understand `database_exists?` method is necessary to see if users need to execute `rails db:create` or not.

Unlike other mysql2 and postgresql adapters, Oracle enhanced adapter `db:create` task creates a database user (someone caii it a database schema), not "Oracle database" since the definition of database are different between Oracle and others.

https://github.com/rsim/oracle-enhanced/blob/c0f24b4668ad45b6b320b8199681fba0c57483d6/lib/active_record/connection_adapters/oracle_enhanced/database_tasks.rb#L15-L29

Then `database_exists?` needs to return if the database user(schema) exists or not, then I think one of the candidates error message should be: `ORA-01017: invalid username/password; logon denied`.

Unfortunately this error message could be reported:
- When the database user does not exist
- When the database user exists but the password is incorrect

This error message can be used to raise `NoDatabaseError` for the first case only, for the second one the database (Here I mean database schema in Oracle) exists but the password is incorrect. By handling all of ORA-01017 Oracle error message as `NoDatabaseError` exception, which cause users misunderstanding the cause of this `NoDatabaseError` like setting incorrect password in database.yml .

koic:
Thank you for your investigation. I opened this PR in preparation for PR #1906. However, I will close this PR because Oracle database cannot detect `NoDatabaseError`. Thank you again.

PR #1923, opened 2019-09-04
Disable `Style/BracesAroundHashParameters` cop

koic:
Follow up https://github.com/rails/rails/pull/37121.


yahonda:
Good catch. Thanks.

PR #1931, opened 2019-09-12
Enabled Dependabot by creating .dependabot/config.yml

yahonda:
This pull request enables Dependabot to see if it works and helps Oracle enhanced adapter development.

Refer https://dependabot.com/docs/config-file/ to create with only required options.


yahonda:
Let's see how it goes. Especially Oracle enhanced adapter does not push `Gemfile.lock` then not sure if it actually helps.

PR #1938, opened 2019-09-14
Disable Code Climate

yahonda:
This pull request disables Code Climate because of:

* Code Climate was enabled at #1057 to run RuboCop
* RuboCop is running using GitHub Actions by #1925 now
* We are not making use of other Code Climate features
* I prefer not to run a lot of jobs at CI

Thanks to Code Climate for providing great services so far.


yahonda:
It looks like just removing `.codeclimate.yml` introduces the default rule of Code Climate.
Let me take another look at "83 issues" reported.

PR #1951, opened 2019-11-08
Test disabling of CodeClimate

rsim:
after disabling it on the CodeClimate web page


yahonda:
Thank you for taking care of it. It looks Code Climate check is not running.

yahonda:
Let me merge #1938 because it explains the reason why. Thanks again for the maintenance.

PR #1988, opened 2020-02-12
Include sync option when dumping a context index

### LINKED ISSUE #1987: Schema dumper does not dump `sync` option
### Steps to reproduce

```ruby
# active_record_gem_spec.rb

ActiveRecord::Base.logger = Logger.new(STDOUT)

describe "bug test" do
  before(:all) do
    ActiveRecord::Base.establish_connection(CONNECTION_PARAMS)
    ActiveRecord::Schema.define do
      create_table :posts, force: true do |t|
        t.string :title
        t.string :body
        # If I comment the `t.index :title` out, the test passes
        t.index :title
      end
      add_context_index :posts, :body, sync: "ON COMMIT"
    end
    class Post < ActiveRecord::Base
    end
  end

  it "should dump the context index" do
    stream = StringIO.new
    ActiveRecord::SchemaDumper.dump(ActiveRecord::Base.connection, stream)
    expect(stream.string).to include(%(sync: "ON COMMIT"))
  end
end
```

### Expected behavior
The dumped index should have a `sync` option

### Actual behavior

It dumps the context index, but it's missing the `sync` configuration.

### System configuration
**Rails version**: 5.2

**Oracle enhanced adapter version**: master and 5.2.8

**Ruby version**: 2.5.1

**Oracle Database version**: 12.2.0.1

jhanggi:
Includes `sync` option in `schema.rb` when dumping a context index.

Fixes #1987 

I didn't want to duplicate the test case I wrote in #1986, so this also includes that. I'm happy to split it out if you're rather see it separately.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_dumper.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_dumper.rb
@@ -50,6 +50,7 @@ module ActiveRecord #:nodoc:
                   else
                     statement_parts = [ ("add_context_index " + remove_prefix_and_suffix(table).inspect) ]
                     statement_parts << index.columns.inspect
+                    statement_parts << ("sync: " + $1.inspect) if index.parameters =~ /SYNC\((.*?)\)/
                     statement_parts << ("name: " + index.name.inspect)
                   end
                 else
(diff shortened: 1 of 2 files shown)


aiomaster:
I can confirm these changes work for me and fix the described bug.

jhanggi:
Hi @yahonda is it possible to get this and #1986 backported to the 5.2 branch? What's the best way to do that? Should I issue a new PR?

yahonda:
Thanks for the comment.

I usually maintain and release the latest version of Oracle enhanced adapter.
As of today, It is Oracle enhanced adapter 6.0.x. No older version will be released.

Although this may not be the best option for you, you can backport #1986 to `release52` branch at your fork by yourself. This is what I can say now.

Thank you for your understanding.

jhanggi:
Understood. Thanks for responding. We'rer still at Rails 5.2 so we can't use the current adapter version. Hopefully we'll get up to Rails 6.0 on our side at some point soon and get back at the current release.

PR #1994, opened 2020-03-16
Use `build_result` instead of `ActiveRecord::Result.new`

koic:
Follow https://github.com/rails/rails/pull/38713.


yahonda:
Thank you!

PR #2001, opened 2020-04-14
[WIP] Migrate Arel code from Rails to Oracle enhanced adapter

yahonda:
This is a draft pull request to support https://github.com/rails/rails/pull/38946 

### Required
- [ ] Migrate Arel library code to make CI green
- [ ] Migrate git commit history

### Optional
- [ ] Migrate test code
because the unit test for Arel is written in Minitest, Oracle enhanced adapter uses RSpec. Arel has been a private API at Rails then it may be fine without tests for the time being.

PR #2002, opened 2020-04-15
Merge Arel visitors

koic:
Follow up of rails/rails#38946 and rsim/oracle-enhanced#2001.

This PR merges Arel visitors for Oracle to Active Record Oracle enhanced adapter. It includes visitors tests written in `Arel::Spec (Minitest::Spec)`, so it should be migrated to RSpec in future.


yahonda:
Thanks for the great work.

koic:
Thank you too, always!

PR #2007, opened 2020-04-21
Revert part of #2006

koic:
This PR removes the wrong diff on #2006.
This diff is an upstream change made in https://github.com/rails/rails/commit/36ea108.

This PR reverts for a different commit reason than #2006.


koic:
I will open a new PR for this upstream follow-up when confirmation is completed.

PR #2008, opened 2020-04-26
Import `visit_Arel_Nodes_{In,NotIn}` from Rails code base

kamipo:
`connection.in_clause_length` and related code will be removed from
Rails code base, so should be owned in the repo itself.

Ref https://github.com/rails/rails/pull/39057.


yahonda:
Thanks!

PR #2013, opened 2020-05-08
Limit number of expressions in a list during a "homogenous in" operation

jamesnakagawa:
* Oracle limits lists to a maximum of 1,000 items. 
* rails/rails@72fd0bae5948c1169411941aeea6fef4c58f34a9 implements a new type of `in` operation which can be performed under certain circumstances.
* This caused the test "OracleEnhancedAdapter eager loading should load included association with more than 1000 records" to fail with ORA-01795.
* The solution was to implement `visit_Arel_Nodes_HomogeneousIn`.
* Fixed a typo while I was at it.


yahonda:
Thank you for your contribution. Going forward, it would be appreciated if you open another pull request for fixing a typo.

PR #2016, opened 2020-05-15
Type casted binds fix 5 0 7 x

loop23:
Hello,
ActiveRecord 5.0.7x introduced  a change in abstract_adapter log method, where now it takes type_casted_binds as fourth argument.
So just adding it (you already calculate it for the block) in method call seems to work.
I'm not messing with versions etc, you decide if wanna release 1.17.12, also because you would have to update readme.
Also, I'm not convinced my pull request is correct: If in github I select base repository with yours, and base to be the v1.17.11 tag it does not let me create pull request, it only does if base is master. 

But of course you want to just apply the edits to tag v1.17.11

Bye


yahonda:
Thanks for opening a pull request. I only maintain the latest version of Oracle enhanced adapter for Rails 6.0. I understand some of users still need to use older versions of Rails, but this is what it is.

Technically, there are some conflicts because this pull request points to the master. You can cherry-pick commits into `release17` branch at your repository.

Thanks again for the pull request.

PR #2019, opened 2020-05-24
Follow rails/rails#39365

koic:
Follow up to rails/rails#39365.

This PR consists of the following two commits.

1. Default engine `ENGINE=InnoDB` is no longer dumped to make schema more agnostic (rails/rails@3c6be5e)
2. Separate primary key column options from table options (rails/rails@859681e)

The latter changes `create_table` format in the dumped schema file.

PR #2047, opened 2020-09-10
Restore Schema Dumper behavior changed by #2019

yahonda:
This pull request restores Schema Dumper behavior changed by #2019
because rails/rails#39365 does not intend to dump the default primary key configuration.

Related to rails/rails#39365
Follow up #2019

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_dumper.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_dumper.rb
@@ -4,7 +4,16 @@ module ActiveRecord #:nodoc:
   module ConnectionAdapters #:nodoc:
     module OracleEnhanced #:nodoc:
       class SchemaDumper < ConnectionAdapters::SchemaDumper #:nodoc:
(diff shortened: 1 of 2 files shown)


lib/active_record/connection_adapters/oracle_enhanced/schema_dumper.rb
```diff
@@ -5,6 +5,12 @@ module ConnectionAdapters #:nodoc:
     module OracleEnhanced #:nodoc:
       class SchemaDumper < ConnectionAdapters::SchemaDumper #:nodoc:
         private
+          def column_spec_for_primary_key(column)
+            spec = super
+            spec.except!(:precision) if prepare_column_options(column) == { precision: "38", null: "false" }
```
koic, line 10: I'm not sure if the comparison with the hash object is the best way, but the dumped source has been fixed.
I have a one point. I think that the hash object instantiation can be reduced by making the hash object to be a constant.

yahonda:
Thanks for the review. Updated to save the frozen hash as a constant.

koic (approved):
Thank you for removing weird dumped table option!

PR #2052, opened 2020-10-03
CI against Ruby 2.7.2

koic:
Ruby 2.7.2 has been released.
https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ script:
 
 language: ruby
 rvm:
-  - 2.7.1
+  - 2.7.2
   - 2.6.6
   - 2.5.8
   - jruby-9.2.13.0


yahonda:
Thank you. Do you think is there anything we need to configure to enable warnings which have been disabled in Ruby 2.7.2 by default?

yahonda:
I should have been more clear.

https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/ says if we want to enable warnings use these options.

> You can turn on the deprecated warnings by specifing command line option -w or -W:deprecated. 

I'm wondering if we need to change command line option for `bundle exec rake spec` 

https://travis-ci.org/github/rsim/oracle-enhanced/jobs/732540066#L457

Let me take some more look at how to enable warnings for RSpec and minitest.

koic:
That's a good point. So, the incompatibility warning from Ruby 2.7 to Ruby 3.0 causes an error in ruby-head (Ruby 3.0.0-dev).
In other words, Ruby 2.7 incompatibility warnings can be detected as errors in the ruby-head matrix, so I think it's unnecessary to specify `-W:deprecated` for Ruby 2.7.2.

yahonda:
That makes sense. Thank you.

PR #2054, opened 2020-10-08
Add missing method to Oracle columns

coorasse:
This is currently available for Postgres https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql/column.rb#L21.
can we have it as well? does oracle enhanced support array columns?

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/column.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/column.rb
@@ -13,6 +13,10 @@ module ActiveRecord
         def virtual?
           virtual
         end
+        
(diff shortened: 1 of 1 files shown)


yahonda:
Thank you for opening a pull request.

> does oracle enhanced support array columns?
As far as I know, PL/SQL has some collection data types https://docs.oracle.com/cd/B28359_01/appdev.111/b28371/adobjcol.htm#ADOBJ00210 but I do not know if any array columns supported by Oracle database column type.

Would you mind providing some use cases?

yahonda:
Closing this pull request.

PR #2070, opened 2020-11-19
`build_subselect` does not have ordering

koic:
Fixes #2023.

This PR fixes `ActiveRecord::StatementInvalid: OCIError: ORA-00907: missing right parenthesis` error when using assoc has `dependent: :delete_all` and `order`.

```ruby
has_many :assoc, -> { order(:foo) }, dependent: :delete_all
```

Oracle does not accept `ORDER BY` in `DELETE` and `UPDATE`'s subquery. This PR adds the following reproduction test and passes it.

```console
vagrant@rails-dev-box:~/src/oracle-enhanced$ bundle exec rspec
spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb:187
==> Loading config from ENV or use default
==> Running specs with ruby version 2.7.2
==> Effective ActiveRecord version 6.1.0.rc1
Run options: include
{:locations=>{"./spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb"=>[187]}}
F

Failures:

  1) OracleEnhancedAdapter `has_many` assoc has `dependent: :delete_all`
  with `order` should load included association with more than 1000
  records
     Failure/Error: @raw_cursor.exec

     ActiveRecord::StatementInvalid: OCIError: ORA-00907: missing right parenthesis
     # stmt.c:267:in oci8lib_270.so
     # /home/vagrant/.rvm/gems/ruby-2.7.2/bundler/gems/ruby-oci8-446ad166808c/lib/oci8/cursor.rb:137:in `exec'
     # ./lib/active_record/connection_adapters/oracle_enhanced/oci_connection.rb:150:in `exec_update'
     # ./lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb:158:in `block (2 levels) in exec_update'
```

PR #2073, opened 2020-11-30
Merge pull request #2070 from koic/build_subselect_doesnt_have_ordering

yahonda:
This pull request backports #2070 into the release60 branch. 
Since #2002 has been merged to release61 branch or newer, #2070 is not feasible to merge to release60 branch. Then applied some dirty monkey patch to `lib/active_record/connection_adapters/oracle_enhanced_adapter.rb` file.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -746,3 +746,26 @@ require "active_record/connection_adapters/oracle_enhanced/version"
 module ActiveRecord
   autoload :OracleEnhancedProcedures, "active_record/connection_adapters/oracle_enhanced/procedures"
 end
+
+# Backport #2070 into relese60 branch by adding some dirty monkey patch
+# Because #2002 has been merged to release61 branch or newer, not available for release60 branch.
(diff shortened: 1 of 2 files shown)


yahonda:
@koic Would you review this pull request?  I do not think this "monkey patch" is the best way but do not want to add new file structures to the release60 branch.

koic (approved):
Thank you for creating the patch. I think this monkey patch approach is not bad.

The following is a comparison with opening a patch to Arel Visitor of Rails repo vs this PR approach.

Pros:

- Oracle enhanced adapter gem can have a release cycle independent of Rails.
- No need to have Rails maintainers consider Oracle-specific issue.

Cons:

- Arel Visitor of Oracle packaged in the upstream Rails 6.0 series never resolve this issue.

Also, Rails 6.0 series is a stable version. So I think any drastic changes will not be made to Arel Visitor, so I think it's unnecessary to copy all of Arel Visitor.

Conclusion. In practice, Active Record Oracle enhanced apapter for Rails 6.0 should also be updated to the latest version if user expect bug fixes. So I'm not sure whether this monkey patch is the best option, but I think it's a better option.

yahonda:
Thanks for the review.

PR #2076, opened 2020-12-14
Usage of bind variables for volatile filter conditions

rammpeter:


rammpeter:
Sorry for the inconvenience from my side: "JDBC statement cache size" has been planned as a separate pull request but the commit shows up now in the current pending pull request for "usage of binds".

This extension allows to keep SQL cursors open against the DB even if AR closes them after statement execution.
This way effort for soft parses in DB can be dramatically reduced.

yahonda:
Thanks for the pull request. Would you open another pull request for `config[:jdbc_statement_cache_size]`?

rammpeter:
Recreating pull request with according commits

PR #2083, opened 2020-12-15
Fix write_lobs "invalid byte sequence in UTF-8"

yegayazilim:
write_lobs checks to see whether the value is blank before executing the SQL statement. If the LOB is binary blank? call causes "invalid byte sequence in UTF-8" exception. We can instead use unless value, which is applicable both to strings and binary data.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb
@@ -257,6 +257,8 @@ module ActiveRecord
             next unless value
             if klass.attribute_types[col.name].is_a? Type::Serialized
               value = klass.attribute_types[col.name].serialize(value)
+              # value can be nil after serialization because ActiveRecord serializes [] and {} as nil
+              next unless value
             end
             uncached do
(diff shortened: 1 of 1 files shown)


yahonda:
Thank you for opening a pull request. It looks like the CI fails. Would you address this CI failure and add some RSpec test which makes sure what this pull request fixes.  Thanks.

yegayazilim:
I tried to write an RSpec test for the condition I'm fixing, but I couldn't. It happens when I try to save an .xlsx document to a binary column in Oracle.  write_lobs method uses value.blank? on this binary object and blank? throws "invalid byte sequence in UTF-8". So, I'm trying to avoid using blank? in the method. CI was failing in serialized column test case. In that test, value is serialized in write_lobs method. After serialization value can be nil because ActiveRecord persists empty Array and Hash as null in the database. So, I added another nil check after serialization to fix the CI problem. I also added a comment

yahonda:
CI for existing specs is green now. Would you add some specs to make sure which issue is fixed by this pull request? Thank you.

yahonda:
Closed via #2097

Issue #1678, opened 2018-03-21
Missing IN or OUT parameter in 5.2.0.rc1


remofritzsche:
Sorry to bother you once more, but after updating to 5.2.0.rc1, we're experiencing the issue as described in #1070 again. The setting `use_old_oracle_visitor` is still set to `true`. Thanks a lot for having a look at this 👍 

```
ActiveRecord::StatementInvalid (Java::JavaSql::SQLException: Missing IN or OUT parameter at index:: 1: select us.sequence_name from all_sequences us where us.sequence_owner = :owner and us.sequence_name = :sequence_name)
```

### System configuration
**Rails version**: 5.2.0.rc2

**Oracle enhanced adapter version**: 5.2.0.rc1

**Oracle JDBC driver version**: 12.1.0.2 (com.oracle.jdbc / ojdbc7)

**Ruby version**: jruby 9.1.14.0 (2.3.3)

**Oracle Database version**:  12.1.0.2.0

yahonda:
Thanks for the report.

I recall the method `combine_bind_parameters ` changed by #1033 has been [removed from Rails](https://github.com/rails/rails/commit/213796fb4936dce1da2f0c097a054e1af5c25c2c) and [Oracle enhanced adapter](https://github.com/rsim/oracle-enhanced/pull/1527). There is a related change at [Arel](https://github.com/rails/arel/pull/491)

Let me take some more look at this. I may be going to ask you to provide an executable test case to save some time and to validate we are working on the same issue.

Thanks.

remofritzsche:
Thanks for your quick response. I'll try and create an executable test case for you so it's easier to reproduce.

remofritzsche:
I'm happy to say that I can finally reproduce it in an isolated test :) Please see https://gist.github.com/remofritzsche/76f84c68f5545fbb46e50f6d3b633884 for a minimal example test case. Thanks a lot for having a look at this.

yahonda:
Thanks for providing the script. 

It reproduces `ActiveRecord::StatementInvalid: Java::JavaSql::SQLException: Missing IN or OUT parameter at index:: 1:` exception as follows. 

I would like you to confirm the sql statement is same as yours `select us.sequence_name from all_sequences us where us.sequence_owner = :owner and us.sequence_name = :sequence_name`

```ruby
  1) Error:
BugTest#test_3_join_error:
[...]

PR #2087, opened 2020-12-16
Usage of bind variables for volatile filter conditions (2)

rammpeter:

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -562,7 +562,7 @@ module ActiveRecord
       def column_definitions(table_name)
         (owner, desc_table_name) = @connection.describe(table_name)
 
-        select_all(<<~SQL.squish, "SCHEMA")
+        select_all(<<~SQL.squish, "SCHEMA", [bind_string("owner", owner), bind_string("table_name", desc_table_name)])
           SELECT /*+ OPTIMIZER_FEATURES_ENABLE('11.2.0.2') */ cols.column_name AS name, cols.data_type AS sql_type,
                  cols.data_default, cols.nullable, cols.virtual_column, cols.hidden_column,
                  cols.data_type_owner AS sql_type_owner,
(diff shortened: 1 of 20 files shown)


yahonda:
Thanks for the pull request.
`column_definitions` and `pk_and_sequence_for` methods uses SQL literals intentionally since #1713 to fix #1678.
Would you restore original behavior for these methods? Thank you.

rammpeter:
Excuse me please for inconvenience in this case. 
If it would be ok for you, I‘d add a test case first for the problem described in #1678 to raise a test error under current pull request state and fix this than.

yahonda:
> I‘d add a test case first for the problem described in #1678 to raise a test error under current pull request state and fix this than.

Sounds good. As long as the issue described in #1678 is fixed with bind variables it should be fine.

rammpeter:
Issue described in #1678 is fixed now with bind variables.

lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
```diff
@@ -763,6 +763,23 @@ def bind_string(name, value)
           ActiveRecord::Relation::QueryAttribute.new(name, value, Type::OracleEnhanced::String.new)
         end
 
+        # call select_values using binds even if surrounding SQL preparation/execution is done
+        # with conn.unprepared_statement (like AR.to_sql)
+        def select_values_forcing_binds(arel, name, binds)
+          # remove possible force of unprepared SQL during dictionary access
+          unprepared_statement_forced = prepared_statements_disabled_cache.include?(object_id)
```
yahonda, line 770: I did not know `prepared_statements_disabled_cache` method implemented via  https://github.com/rails/rails/commit/206345b9890

lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
```diff
@@ -491,7 +491,7 @@ def prefetch_primary_key?(table_name = nil)
         do_not_prefetch = @do_not_prefetch_primary_key[table_name]
         if do_not_prefetch.nil?
           owner, desc_table_name = @connection.describe(table_name)
-          @do_not_prefetch_primary_key [table_name] = do_not_prefetch = !has_primary_key?(table_name, owner, desc_table_name)
+          @do_not_prefetch_primary_key[table_name] = do_not_prefetch = !has_primary_key?(table_name, owner, desc_table_name)
```
yahonda, line 494: Would you rebase this pull request based on the current master branch? because https://github.com/rsim/oracle-enhanced/pull/2108 addresses this unnecessary spece.

rammpeter:
Sorry, I'm not sure enough if everything worked well with rebase.
I close this pull request and reopen it again based on current head.

PR #2088, opened 2020-12-16
Allow usage of JDBC statement caching by DB config parameter

rammpeter:
This extension allows to keep SQL cursors open against the DB even if AR closes them after statement execution.
This way effort for soft parses in DB can be dramatically reduced.
It is controlled by the DB config parameter "jdbc_statement_cache_size".
Per default statement caching is disabled. This way preserving the previous behavior.


yahonda:
Thank you.

PR #2099, opened 2021-01-04
Backport Fix write_lobs Invalid byte sequence in UTF-8 to release60

yegayazilim:
Backport fix from master.
write_lobs checks to see whether the value is blank before executing the SQL statement. If the LOB is binary blank? call causes "invalid byte sequence in UTF-8" exception. We can instead use unless value, which is applicable both to strings and binary data.


yahonda:
Thanks for opening a backport request. Usually, backport is taken care of by me by cherry-pick merge commit into a specific branch if necessary.  I have backported #2097 to release61 branch via #2111 and to release60 branch via #2112.

PR #2098, opened 2021-01-01
Require `rspec/autorun` to run one line spec file

yahonda:
* Without this commit

```ruby
$ ruby active_record_gem_spec.rb
Fetching https://github.com/rails/rails.git
... snip ...
Using rails 6.2.0.alpha from https://github.com/rails/rails.git (at master@afc79e3)
$
```

* With this commit

```ruby
$ ruby active_record_gem_spec.rb
Fetching https://github.com/rails/rails.git
... snip ...
Using rails 6.2.0.alpha from https://github.com/rails/rails.git (at master@afc79e3)
-- create_table(:posts, {:force=>true})
D, [2021-01-01T17:27:35.727438 #554514] DEBUG -- :    (20.9ms)  DROP TABLE "POSTS"
D, [2021-01-01T17:27:35.733266 #554514] DEBUG -- :    (5.4ms)  DROP SEQUENCE "POSTS_SEQ"
D, [2021-01-01T17:27:35.748060 #554514] DEBUG -- :    (14.4ms)  CREATE TABLE "POSTS" ("ID" NUMBER(38) NOT NULL PRIMARY KEY)
D, [2021-01-01T17:27:35.751081 #554514] DEBUG -- :    (2.8ms)  CREATE SEQUENCE "POSTS_SEQ" START WITH 1
   -> 0.1797s
-- create_table(:comments, {:force=>true})
D, [2021-01-01T17:27:35.869071 #554514] DEBUG -- :    (21.4ms)  DROP TABLE "COMMENTS"
D, [2021-01-01T17:27:35.875660 #554514] DEBUG -- :    (6.3ms)  DROP SEQUENCE "COMMENTS_SEQ"
D, [2021-01-01T17:27:35.884094 #554514] DEBUG -- :    (8.0ms)  CREATE TABLE "COMMENTS" ("ID" NUMBER(38) NOT NULL PRIMARY KEY, "POST_ID" NUMBER(38))
D, [2021-01-01T17:27:35.887143 #554514] DEBUG -- :    (2.8ms)  CREATE SEQUENCE "COMMENTS_SEQ" START WITH 1
   -> 0.1359s
[...]

Issue #2100, opened 2021-01-04
Test failed for pull requests at step "run bug report templates"


rammpeter:
### Steps to reproduce
Retry test of pull request  #2087

### Expected behavior
test without errors

### Actual behavior
Currently the Github test for pull request #2087 fails at step "run bug report templates" with:
/opt/hostedtoolcache/Ruby/2.6.6/x64/lib/ruby/gems/2.6.0/bundler/gems/rails-03516a924e71/activesupport/lib/active_support/dependencies.rb:332:in `require': cannot load such file -- rspec/core/formatters/progress_formatter (LoadError)

Can this possibly be caused by commit e3c59d6 (Require `rspec/autorun` to run one line spec file) of file 
oracle-enhanced/guides/bug_report_templates/active_record_gem_spec.rb
that changed 'require "rspec"' to 'require "rspec/autorun"' ?

yahonda:
#2102 has been merged to master branch then CI should work.

rammpeter:
pull request checks are successful

PR #2102, opened 2021-01-05
Revert "Require `rspec/autorun` to run one line spec file"

yahonda:
Reverts rsim/oracle-enhanced#2098


yahonda:
Revert #2098 for the time being to workaround #2100. #2101 has been opened to address it but actually, it does not yet.

Issue #1625, opened 2018-01-10
Extract views to db/structure.sql by default


yahonda:
### Steps to reproduce
There are some questions about extracting views to db/structure.sql with `rake db:structure:dump`
Refer #1339 #1355 #1624

### Expected behavior
Views are triggered to db/structure.sql by default.
Not decided to have an option not to export views. 

### Actual behavior
To extract views `structure_dump: db_stored_code` needs configured at `database.yml`

### System configuration
**Rails version**:next major version of Rails 6 or later

**Oracle enhanced adapter version**:next major version of Rails 6 or later

**Ruby version**:Ruby version which supported by the next major version of Rails

**Oracle Database version**:Oracle Database 11.2 or later

PR #2109, opened 2021-01-11
Ensure FKs are properly included in structure dumps

joshpencheon:
# Issue Summary

FK constraints are mistakenly not being included in SQL structure dumps (`rails db:structure:dump`).

## Issue Details

Due to a bug that I believe was introduced quite a while ago (in #1641, fixing #1625), the DDL describing foreign keys is being discarded in favour of the DDL describing views.

I believe this was missed because the test coverage of the FK dumping tests only the standalone method that does this (`#structure_dump_fk_constraints`), rather than the main `#structure_dump` method.

# PR Details

This PR ensures that both the view DDL and the foreign key DDL gets included when a structure dump is performed, and adjusts the specs accordingly.

## Note on testing

I could only get a limited Oracle enviroment working, but was able to run the appropriate specs:

```
$ bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/structure_dump_spec.rb
==> Loading config from /Users/josh/work/forks/oracle-enhanced/spec/spec_config.yaml
==> Running specs with ruby version 2.6.3

==> Effective ActiveRecord version 6.2.0.alpha
..............................*..

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) OracleEnhancedAdapter structure dump schema migrations multi insert is NOT supported should dump schema migrations one version per insert
     # Not supported in this database version
     # ./spec/active_record/connection_adapters/oracle_enhanced/structure_dump_spec.rb:376


Finished in 34.71 seconds (files took 1.16 seconds to load)
33 examples, 0 failures, 1 pending
```

Many thanks for all the hard work on this gem! 😃

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/structure_dump.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/structure_dump.rb
@@ -55,8 +55,9 @@ module ActiveRecord #:nodoc:
             structure << structure_dump_column_comments(table_name)
           end
 
-          join_with_statement_token(structure) << structure_dump_fk_constraints
-          join_with_statement_token(structure) << structure_dump_views
+          join_with_statement_token(structure) <<
+            structure_dump_fk_constraints <<
+            structure_dump_views
         end
 
         def structure_dump_column(column) #:nodoc:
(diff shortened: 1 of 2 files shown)


yahonda:
Thank you for the pull request.

PR #2121, opened 2021-01-13
structure dump: read column comments from all_tab_cols

joshpencheon:
## Issue Summary

Currently, when dumping column comments, `user_tab_columns` is used. This misses comments which are in the current schema, when the current user is not the owner.

## PR Summary

This PR dumps column comments from `all_tab_cols` instead, filtering `owner` to the `current_schema`. This is in line with the approach taken for the rest of the objects extracting during the structure dump.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/structure_dump.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/structure_dump.rb
@@ -175,8 +175,9 @@ module ActiveRecord #:nodoc:
         def structure_dump_column_comments(table_name)
           comments = []
           columns = select_values(<<~SQL.squish, "SCHEMA")
-            SELECT /*+ OPTIMIZER_FEATURES_ENABLE('11.2.0.2') */ column_name FROM user_tab_columns
-            WHERE table_name = '#{table_name}' ORDER BY column_id
+            SELECT /*+ OPTIMIZER_FEATURES_ENABLE('11.2.0.2') */ column_name FROM all_tab_columns
+            WHERE owner = SYS_CONTEXT('userenv', 'current_schema')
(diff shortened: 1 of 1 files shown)


joshpencheon:
(The jruby builds appear to be failing for what seems to be unrelated reasons.)

lib/active_record/connection_adapters/oracle_enhanced/structure_dump.rb
```diff
@@ -175,8 +175,9 @@ def structure_dump_table_comments(table_name)
         def structure_dump_column_comments(table_name)
           comments = []
           columns = select_values(<<~SQL.squish, "SCHEMA")
-            SELECT /*+ OPTIMIZER_FEATURES_ENABLE('11.2.0.2') */ column_name FROM user_tab_columns
-            WHERE table_name = '#{table_name}' ORDER BY column_id
+            SELECT /*+ OPTIMIZER_FEATURES_ENABLE('11.2.0.2') */ column_name FROM all_tab_cols
```
yahonda, line 178: Would you explain why `all_tab_cols` is used not `all_tab_columns`?

joshpencheon: That's my ignorance, sorry! Having read [some docs](https://docs.oracle.com/en/database/oracle/oracle-database/12.2/refrn/ALL_TAB_COLS.html#GUID-85036F42-140A-406B-BE11-0AC49A00DBA3) I will swap this out :)

joshpencheon: I've force-pushed with this swapped out, thanks for the feedback. 👍

yahonda:
Thank you.

Issue #2118, opened 2021-01-13
JRuby shows "class org.jruby.ir.operands.TemporaryClosureVariable cannot be cast "


yahonda:
### Steps to reproduce
CI against JRuby fails at https://travis-ci.com/github/rsim/oracle-enhanced/jobs/471296127#L517

### Expected behavior
It should pass.

### Actual behavior
https://travis-ci.com/github/rsim/oracle-enhanced/jobs/471296127#L517

### System configuration
**Rails version**: master

**Oracle enhanced adapter version**: master

**Ruby version**: JRuby 9.1.14.0 and jruby-head

**Oracle Database version**: Oracle Database 18c XE ad 11g XE

yahonda:
Closing it as it does not look like reproduce anymore.

PR #2125, opened 2021-01-15
Usage of bind variables for volatile filter conditions (3)

rammpeter:


rammpeter:
Is there a known issue in current master for jRuby?
> OracleEnhancedConnection default_timezone should respect default_timezone = :utc than time_zone setting
Is raised independent from this pull request.

yahonda:
I am aware of this failure https://github.com/rsim/oracle-enhanced/issues/2118 but did not have time to tackle it yet.

yahonda:
Thanks for your contribution.

PR #2128, opened 2021-01-19
Use longer identifier for `DatabaseLimits` when using Oracle 12.2 or higher

koic:
Follow #1703.

Oracle enhanced adapter supports longer identifier by #1703.
I encountered the following error when using `rename_table` in Oracle 12c.

```console
New table name 'identifier_of_thirty_bytes_or_more' is too long; the limit is 30 characters
```

Because `IDENTIFIER_MAX_LENGTH` constant was fixed to 30 bytes.

This PR will use `max_identifier_length` method instead of the constant and accept longer identifier (max 128 bytes) when using Oracle 12.2 or higher.

I encountered the error in Rails 6.0. So I'd like to backport to 6.1 and 6.0 stable branches.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
@@ -457,7 +457,7 @@ module ActiveRecord
         end
 
         # To allow private method called from `JDBCConnection`
-        def describe(name)
+        def describe(name, supports_longer_identifier)
           super
         end
 
--- a/lib/active_record/connection_adapters/oracle_enhanced/oci_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/oci_connection.rb
@@ -225,7 +225,7 @@ module ActiveRecord
           lob.write value
         end
 
-        def describe(name)
+        def describe(name, supports_longer_identifier)
           super
         end
 
--- a/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
@@ -20,14 +20,14 @@ module ActiveRecord
 
         private
           # Used always by JDBC connection as well by OCI connection when describing tables over database link
-          def describe(name)
+          def describe(name, supports_longer_identifier)
             name = name.to_s
             if name.include?("@")
               raise ArgumentError "db link is not supported"
             else
               default_owner = @owner
             end
-            real_name = OracleEnhanced::Quoting.valid_table_name?(name) ? name.upcase : name
+            real_name = OracleEnhanced::Quoting.valid_table_name?(name, supports_longer_identifier) ? name.upcase : name
             if real_name.include?(".")
               table_owner, table_name = real_name.split(".")
             else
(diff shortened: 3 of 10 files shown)


lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
```diff
@@ -457,7 +457,7 @@ def write_lob(lob, value, is_binary = false)
         end
 
         # To allow private method called from `JDBCConnection`
-        def describe(name)
+        def describe(name, supports_longer_identifier)
```
yahonda, line 460: Do you mean `supports_longer_identifier?`

lib/active_record/connection_adapters/oracle_enhanced/oci_connection.rb
```diff
@@ -225,7 +225,7 @@ def write_lob(lob, value, is_binary = false)
           lob.write value
         end
 
-        def describe(name)
+        def describe(name, supports_longer_identifier)
```
yahonda, line 228: Do you also mean `supports_longer_identifier?`?

yahonda (commented):
Thanks for the pull request. I'm not sure if `supports_longer_identifier` or `supports_longer_identifier?` are good argument names yet. At least this argument name looks long. 

Let me have some more time to consider that.

lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
koic: Yes. It takes the result of `ActiveRecord::ConnectionAdapters::OracleEnhancedAdapter#supports_longer_identifier?` as argument. Unfortunately no better name has come up yet.

koic:
I found that there will probably be breaking change in this PR. Please wait for a while due to I will investigate it.

yahonda:
Sure.

Although I have not checked this pull request in detail yet, In general, I prefer not to change method signature/arity for this requirement. Because the Oracle database version never changes per database connection. Also, I especially prefer not to change the method signature if it requires to backport to the older versions of Oracle enhanced adapter.

Let's have some discussion if you have questions. Thank you.

pasku:
Any updates on this?

mattalat:
Is there any work that needs to be completed on this? If so, I'd be happy to help. I am eager to see this in a release!

akostadinov:
If Rails can create proper relationships with the new column type, then I would say, use it by default unless user requested older style primary key. Or course some test cases to show that would be needed.

akostadinov:
The biggest complaint I see is changing of the method signatures. Since identifiers are same for the whole connection, I think that probably detecting their length within the connection and then accessing that value wherever needed would resolve that.

Also my suggestion would be to have a method that returns length of identifiers instead of `long`/`short` and other non-descriptive names. This may allow users to set some configuration option to force certain identifiers max length.

JackieSL1:
Any updates on this?

akostadinov:
Another issue I see is that once the identifiers are switched to a different length, the table sequences for auto-incrementing primary keys may change names for long table names.

At least previously when tables were created with an auto-increment primary key (the default), a sequence was created with name `TABLE_NAME_SEQ`. If `TABLE_NAME` is long, then it will be truncated by current implementation.

With the new implementation it will have the full name.

This difference will cause `ActiveRecord::StatementInvalid (OCIError: ORA-02289: sequence does not exist)`

So a migration PATH must be developed to update all necessary sequence names or somehow regenerate all sequences, idk.

All said, it would be awesome to finally have long identifiers by default. So thank you for your contribution.

PR #2140, opened 2021-02-08
Enable `Performance/BindCall` cop

yahonda:
Follow https://github.com/rails/rails/commit/6515d6985d2b537486bd9d1faa95212a508dc42c


koic:
Thank you!

PR #2155, opened 2021-03-17
Support use of ojdbc11.jar

rammpeter:
Oracle provides separate JDBC version for Java 11 and above (ojdbc11.jar).

This change allows application start if only ojdbc11.jar is provided.
It forces usage of ojdbc11.jar instead of ojdbc8.jar if Java 11+ is used and ojdbc11.jar available.


yahonda:
Thanks for opening a pull request. Would you rebase the branch based on the maseter?

yahonda:
Nevermind. CI against JRuby is not running at master branch yet. then merging this pull request for now.

PR #2188, opened 2021-06-16
Address "NoMethodError: undefined method `partial_writes?'"

yahonda:
This commit addresses the following failures since rails/rails#42355 .
`DEPRECATION WARNING` have not been fixed yet.

```ruby
$ bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/procedures_spec.rb
==> Loading config from ENV or use default
==> Running specs with ruby version 3.0.1
==> Effective ActiveRecord version 7.0.0.alpha
..FFDEPRECATION WARNING: `ActiveRecord::Base.partial_writes=` is deprecated and will be removed in Rails 7.1. Use `partial_updates=` and `partial_inserts=` instead. (called from block (2 levels) in <top (required)> at /home/yahonda/src/github.com/rsim/oracle-enhanced/spec/active_record/connection_adapters/oracle_enhanced/procedures_spec.rb:214)
FDEPRECATION WARNING: `ActiveRecord::Base.partial_writes=` is deprecated and will be removed in Rails 7.1. Use `partial_updates=` and `partial_inserts=` instead. (called from block (2 levels) in <top (required)> at /home/yahonda/src/github.com/rsim/oracle-enhanced/spec/active_record/connection_adapters/oracle_enhanced/procedures_spec.rb:227)
F....F.DEPRECATION WARNING: `ActiveRecord::Base.partial_writes=` is deprecated and will be removed in Rails 7.1. Use `partial_updates=` and `partial_inserts=` instead. (called from block (2 levels) in <top (required)> at /home/yahonda/src/github.com/rsim/oracle-enhanced/spec/active_record/connection_adapters/oracle_enhanced/procedures_spec.rb:319)
F...

Failures:

  1) OracleEnhancedAdapter custom methods for create, update and destroy should update record
     Failure/Error: if partial_writes?

     NoMethodError:
       undefined method `partial_writes?' for #<TestEmployee id: 3, first_name: "Second", last_name: "Last", hire_date: "2008-06-28", salary: nil, description: "First Last", version: 1, create_time: "2021-06-16", update_time: "2021-06-16", created_at: nil, updated_at: "2021-06-16 13:25:56.171290000 +0000">
       Did you mean?  partial_updates?
     # /home/yahonda/src/github.com/rails/rails/activemodel/lib/active_model/attribute_methods.rb:521:in `method_missing'
     # ./lib/active_record/connection_adapters/oracle_enhanced/procedures.rb:153:in `block in _update_record'
[...]

PR #2189, opened 2021-06-17
Suppress `partial_writes` deprecation warning

koic:
Follow rails/rails#42355 and rsim/oracle-enhanced#2188.

This PR suppresses `partial_writes` deprecation warning.


yahonda:
Thank you.

PR #2191, opened 2021-06-17
CI against JRuby 9.2.19.0

yahonda:
* JRuby 9.2.19.0 Released
https://www.jruby.org/2021/06/15/jruby-9-2-19-0.html


koic:
Good catch! Should MRI also be updated?

yahonda:
#2192 has been opened to bump CRuby versions.

koic:
Thanks!

PR #2197, opened 2021-07-13
Revert "Suppress DEPRECATION WARNING by setting ActiveRecord::Base.destroy_all_in_batches = true"

yahonda:
Reverts rsim/oracle-enhanced#2194


yahonda:
Refer https://github.com/rails/rails/pull/42763 for the background.

PR #2203, opened 2021-08-02
all attributes serialized before writing

### LINKED ISSUE #2200: #write_lobs should always serialize attributes
[#write_lobs](https://github.com/rsim/oracle-enhanced/blob/release17/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L718) only serializes attribute before writing to the database when it is `ActiveRecord::Type::Serialized`. Then `value.to_s` is written to the database ignoring ActiveRecord type serialization method.

Attributes should always be serialized before writing to the database. I don't see how custom attributes are supposed to be used otherwise.

A workaround for this would be to define `#to_s` of the custom attribute which does the same as serialization to database, but this is not often undesirable.

I tested with 1.7 as stated at the bottom but I see [same logic used in master](https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb#L258) too.

### Steps to reproduce
1. have a CLOB column type
2. have a field with custom attribute, e.g. `class MyCustomAttribute < ActiveRecord::Type::Text`
3. try to use attribute
```
my_model.myattribute = "whatever long serialized string ..."
my_model.save!
my_model.myattribute
=> whatever deserialized object as expected
my_model.reload
my_model.myattribute
```

### Expected behavior
```
=> whatever deserialized object as expected
```
Basically everything works as if `ActiveRecord::Type::Text` has been used.

### Actual behavior
```
=> "#<MyDeserializedClass:0x0000000010967bd8>tring ..."
```

### System configuration
**Rails version**: 5.0.7.2

**Oracle enhanced adapter version**: activerecord-oracle_enhanced-adapter (1.7.11)

**Ruby version**: ruby 2.5.9p229bu

**Oracle Database version**: oracle:19.3.0

akostadinov:
As described in #2200, presently only `ActiveRecord::Type::Serialized`
attributes are serialized before being written to database by `#write_lobs`.

It was introduced with 75ed13e6a7c42024ff4d57642a1a06f2bdfedc99

Any other attributes are just converted to string by `value.to_s`.

My understanding is that all attributes should go through standard serialization regardless their type. Especially any custom attributes.
Otherwise instead of writing whatever serialized form of attribute is, database record would become `"#<MyCustomAttributeType:0x000000000bdb6590>\",\"whaever garbage...` or some other string representation that was not intended to end up inside the database.

With this update all attributes get serilized in the standard way.

/fixes #2200

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb
@@ -255,11 +255,9 @@ module ActiveRecord
             value = attributes[col.name]
             # changed sequence of next two lines - should check if value is nil before converting to yaml
             next unless value
-            if klass.attribute_types[col.name].is_a? Type::Serialized
-              value = klass.attribute_types[col.name].serialize(value)
-              # value can be nil after serialization because ActiveRecord serializes [] and {} as nil
-              next unless value
-            end
+            value = klass.attribute_types[col.name].serialize(value)
+            # value can be nil after serialization because ActiveRecord serializes [] and {} as nil
+            next unless value
             uncached do
               unless lob_record = select_one(sql = <<~SQL.squish, "Writable Large Object")
                 SELECT #{quote_column_name(col.name)} FROM #{quote_table_name(table_name)}
--- a/spec/active_record/oracle_enhanced/type/custom_spec.rb
+++ b/spec/active_record/oracle_enhanced/type/custom_spec.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+describe "OracleEnhancedAdapter custom types handling" do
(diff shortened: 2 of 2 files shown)


yahonda:
Thanks for opening a pull request. Would you add some specs to make sure your change will not be broken in the future? Thank you.

akostadinov:
@yahonda , good idea, I added. If you can, please take a look. Thank you!

P.S. while writing the tests I noticed something that I didn't expect. I don't know if it is a bug or expected behavior. I will appreciate feedback on it as well - #2226

yahonda:
Just a note: `if klass.attribute_types[col.name].is_a? ActiveRecord::Type::Serialized` has been added via https://github.com/rsim/oracle-enhanced/pull/878

PR #2219, opened 2021-12-02
Add `if_exists` option to `remove_index`

koic:
Follow up to https://github.com/rails/rails/pull/38967.

The `if_exists` option has been introduced since Rails 6.1.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -330,6 +330,8 @@ module ActiveRecord
         # Remove the given index from the table.
         # Gives warning if index does not exist
         def remove_index(table_name, column_name = nil, **options) # :nodoc:
+          return if options[:if_exists] && !index_exists?(table_name, column_name, **options)
+
           index_name = index_name_for_remove(table_name, column_name, options)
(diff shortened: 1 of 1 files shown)


yahonda:
At least this pull request needs this test passed. Would you execute it and see if it works?

https://github.com/rails/rails/blob/2b6182cbd68231ed2a26417e3184c3d8ed21c57a/activerecord/test/cases/migration/index_test.rb#L84-L92

koic:
Ah, your suggestion makes sense! However unfortunately it cannot be executed because my Oracle adapter development environment is out of order yet.

yahonda:
Understood. Let me try running this test with this pull request.

yahonda:
I have confirmed this pull request passes the test added via https://github.com/rails/rails/pull/38967

```ruby
$ ARCONN=oracle bin/test test/cases/migration/index_test.rb -n test_remove_index_which_does_not_exist_doesnt_raise_with_option
Using oracle
Run options: -n test_remove_index_which_does_not_exist_doesnt_raise_with_option --seed 16315

# Running:

.

Finished in 2.123981s, 0.4708 runs/s, 0.4708 assertions/s.
1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
$

yahonda (commented):
LGTM

PR #2236, opened 2021-12-26
CI against Ruby 3.1 at Travis CI

yahonda:
* Ruby 3.1.0 Released
https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/


yahonda:
https://github.com/rvm/rvm/pull/5169 may need merged.

PR #2247, opened 2022-01-12
support LIKE case-insensitive matching

akostadinov:
Oracle documentation suggests to use `UPPER()` when case-insensitive search is desired. I have used that here. Without it the comparison is only case-sensitive unlike MySQL and Postgres where it can be one or the other by user choice.

Probably upgrade notes should point out that by default matching is case-insensitive. If anybody used it as case-sensitive without passing the `case_sensitive` parameter to `#matches`, then they can be surprised by this change.

### DIFF
--- a/lib/arel/visitors/oracle.rb
+++ b/lib/arel/visitors/oracle.rb
@@ -1,8 +1,12 @@
 # frozen_string_literal: true
 
+require_relative "oracle_common"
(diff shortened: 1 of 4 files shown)


yahonda:
Note: This change is breaking one so it will not be backported to the release70 branch.

PR #2250, opened 2022-01-19
Fix broken link

swamp09:
docker-oracle-xe-11g has been removed from docker hub.

Change link to docker-oracle-xe-11g-r2
https://github.com/wnameless/docker-oracle-xe-11g


koic:
Good catch!

PR #2258, opened 2022-01-28
support LOB equality

### LINKED ISSUE #2239: Matching for equality of CLOB fields fails
### Steps to reproduce
```sh
# just a minimal project `rails new --database=oracle
git clone https://github.com/akostadinov/rails-ora-02014.git
cd rails-ora-02014
vi config/database.yml # set your DB connection details
rails c
> a = Article.create!(title: "test", body: "test body")
> Article.where(body: "test body").take
```

### Expected behavior
```ruby
  Article Load (31.3ms)  SELECT "ARTICLES".* FROM "ARTICLES" WHERE DBMS_LOB.COMPARE("ARTICLES"."BODY", :a1) = 0 FETCH FIRST :a2 ROWS ONLY  [["body", #<OCI8::CLOB:0x0000000004035328>], ["LIMIT", 1]]                       
=>                                                          
#<Article:0x000000000404a188                                
 id: 1,
 title: "test",
 body: "test body",
 created_at: Fri, 07 Jan 2022 16:46:23.902340000 UTC +00:00,
 updated_at: Fri, 07 Jan 2022 16:46:23.902340000 UTC +00:00>
```

### Actual behavior
```ruby
  Article Load (2.7ms)  SELECT "ARTICLES".* FROM "ARTICLES" WHERE "ARTICLES"."BODY" = :a1 FETCH FIRST :a2 ROWS ONLY  [["body", #<OCI8::CLOB:0x00007fed8cd6fea8>], ["LIMIT", 1]]                                                                 
Traceback (most recent call last):                                              
stmt.c:265:in oci8lib_270.so: ORA-00932: inconsistent datatypes: expected - got CLOB (OCIError)                                                                 
stmt.c:265:in oci8lib_270.so: OCIError: ORA-00932: inconsistent datatypes: expected - got CLOB (ActiveRecord::StatementInvalid)
```

### Possible fix using a function
This is how we monkey patch the issue. I will probably create a pull request later but wanted to dump what I have here first so I don't forget.

```ruby
    Arel::Visitors::Oracle12.class_eval do # same goes for Oracle
      def visit_Arel_Nodes_Equality(o, collector)
        case (left = o.left)
        when Arel::Attributes::Attribute
          table = left.relation.table_name
          schema_cache = @connection.schema_cache

          return super unless schema_cache.data_source_exists?(table)

          column = schema_cache.columns_hash(table)[left.name.to_s]

          case column.type
          when :text, :binary
            # https://docs.oracle.com/cd/B19306_01/appdev.102/b14258/d_lob.htm#i1016668
            # returns 0 when the comparison succeeds
            comparator = Arel::Nodes::NamedFunction.new('DBMS_LOB.COMPARE', [left, o.right])
            collector = visit comparator, collector
            collector << ' = 0'
            collector
          else
            super
          end
        else
          super
        end
      end
    end
```

### System configuration
**Oracle Enhanced Adapter version**
7.0.0

**Rails version**:
Rails 7.0.1

**Ruby version**:
ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x86_64-linux]

**Oracle version**:
19.3.0

akostadinov:
fixes #2239

### DIFF
--- a/lib/arel/visitors/oracle_common.rb
+++ b/lib/arel/visitors/oracle_common.rb
@@ -4,6 +4,21 @@ module Arel # :nodoc: all
   module Visitors
     module OracleCommon
       private
+        # Oracle can't compare CLOB columns with standard SQL operators for comparison.
+        # We need to replace standard equality for text/binary columns to use DBMS_LOB.COMPARE function.
+        # Fixes ORA-00932: inconsistent datatypes: expected - got CLOB
+        def visit_Arel_Nodes_Equality(o, collector)
+          left = o.left
+          return super unless %i(text binary).include?(cached_column_for(left)&.type)
+
+          # https://docs.oracle.com/cd/B19306_01/appdev.102/b14258/d_lob.htm#i1016668
+          # returns 0 when the comparison succeeds
+          comparator = Arel::Nodes::NamedFunction.new("DBMS_LOB.COMPARE", [left, o.right])
+          collector = visit comparator, collector
+          collector << " = 0"
+          collector
+        end
+
         def visit_Arel_Nodes_Matches(o, collector)
           if !o.case_sensitive && o.left && o.right
             o.left = Arel::Nodes::NamedFunction.new("UPPER", [o.left])
@@ -11,6 +26,20 @@ module Arel # :nodoc: all
           end
 
           super o, collector
+        end
+
+        def cached_column_for(attr)
+          return unless Arel::Attributes::Attribute === attr
+
+          table = attr.relation.table_name
+          return unless schema_cache.columns_hash?(table)
+
+          column = attr.name.to_s
+          schema_cache.columns_hash(table)[column]
+        end
+
+        def schema_cache
+          @connection.schema_cache
         end
     end
   end
--- a/spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
+++ b/spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
@@ -648,7 +648,7 @@ describe "OracleEnhancedAdapter" do
       expect(@logger.logged(:debug).last).to match(/\["table_name", "TEST_POSTS"\]/)
     end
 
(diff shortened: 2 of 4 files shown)


lib/arel/visitors/oracle_common.rb
```diff
@@ -4,14 +4,61 @@ module Arel # :nodoc: all
...
+        schema_cache.columns_hash(table)
+      end
+
+      def schema_cache
+        @connection.schema_cache
+      end
+
+      def table_exists? name
+        schema_cache.data_source_exists?(name)
+      end
```
akostadinov, line 61: These are copied off [arel 7.1.4](https://github.com/rails/arel/blob/v7.1.4/lib/arel/visitors/to_sql.rb). These methods were removed from latter arel releases but are needed for `#visit_Arel_Nodes_Equality` to function.

lib/arel/visitors/oracle_common.rb
```diff
@@ -12,6 +27,20 @@ def visit_Arel_Nodes_Matches(o, collector)
...
+          table = attr.relation.table_name
+          return unless schema_cache.columns_hash?(table)
+
+          column = attr.name.to_s
+          schema_cache.columns_hash(table)[column]
+        end
+
+        def schema_cache
+          @connection.schema_cache
+        end
```
akostadinov, line 43: These two methods are mostly logic from [arel 7.1.4](https://github.com/rails/arel/blob/v7.1.4/lib/arel/visitors/to_sql.rb) with the difference that we only obtain cached columns. Otherwise we go into troubles trying to build the cache.

pandu-49:
@yahonda 
Comparing nil against text fields is failing to fetch data as oracle_enhanced started considering text columns as CLOBs.

Example:
Model.where(text_column: nil)

This supposed to fire a query as `SELECT * FROM models WHERE TEXT_COLUMN IS NULL;`. However, in the latest version this is comparing using DBMS_LOB comparision.

Issue #2257, opened 2022-01-28
System user other than "SYSTEM"


ioev:
I noticed in https://github.com/rsim/oracle-enhanced/blob/4ac11553d1e31297b1f7ad421883c9e398e5e9ff/lib/active_record/connection_adapters/oracle_enhanced/database_tasks.rb the system user is hardcoded as "SYSTEM".  We currently use a different username to create tables, so it would be nice if the username could be customized similar to the password.

yahonda:
Thanks for the report. Would you consider opening a pull request?

ioev:
Sure, I can give that a shot.

PR #2261, opened 2022-02-09
Allow overriding system user on db create.

ioev:
#2257

Change to allow the a user to specify a different system user.  I went with only allowing via an environment variable, as adding a new $stdin might quality as a breaking change.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/database_tasks.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/database_tasks.rb
@@ -13,18 +13,14 @@ module ActiveRecord
         end
 
         def create
-          if ENV["ORACLE_SYSTEM_USER"]
-            system_user = ENV["ORACLE_SYSTEM_USER"]
-          else
-            system_user = "SYSTEM"
-            $stdout.puts "Using system user '#{system_user}' (set ORACLE_SYSTEM_USER to override)"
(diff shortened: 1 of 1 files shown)


akostadinov:
Looks consistent with SYSTEM password.

lib/active_record/connection_adapters/oracle_enhanced/database_tasks.rb
```diff
@@ -13,6 +13,13 @@ def initialize(config)
         end
 
         def create
+          if ENV["ORACLE_SYSTEM_USER"]
```
yahonda, line 16: `system_user = ENV["ORACLE_SYSTEM_USER"] || "SYSTEM"` should be fine.

yahonda:
Please rebase from the master branch to address RSpec failures.

yahonda (changes requested):

davidhooey:
Could we get this pull request merged? When connecting to Oracle Cloud Infrastructure databases the SYSTEM user is not available and we are to use the ADMIN user instead. The ability to override this user is the only way we can use this gem when connecting to an Oracle Cloud Autonomous Database.

yahonda:
Thanks for the update. I have opened a new one https://github.com/rsim/oracle-enhanced/pull/2569

PR #2263, opened 2022-02-23
Enable `Performance/OpenStruct` cop

yahonda:
Follow up rails/rails#44473


yahonda:
CI failures should be addressed by #2264

Issue #2226, opened 2021-12-07
#write_lobs called only on update but not on create


akostadinov:
I'm not exactly sure whether this is an issue or expected behavior. I didn't observe any error and I'm not an experienced Oracle dba. It just looks strange.

### Steps to reproduce

run test `should serialize LOBs when creating a record` from #2203 

### Expected behavior

`#write_lobs` is called

```
  TestEmployee Create (77.7ms)  INSERT INTO "TEST_EMPLOYEES" ("FIRST_NAME", "LAST_NAME", "SIGNATURE", "ID") VALUES (:a1, :a2, :a3, :a4)  [["first_name", "Peter"], ["last_name", "Doe"], ["signature", "cGV0ZXInc3NpZ25hdHVyZQ==\n"], ["id", 1]]
  Writable Large Object (2.4ms)  SELECT "SIGNATURE" FROM "TEST_EMPLOYEES" WHERE "ID" = 1 FOR UPDATE
```

### Actual behavior

`#write_lobs` is not called

```
  TestEmployee Create (77.7ms)  INSERT INTO "TEST_EMPLOYEES" ("FIRST_NAME", "LAST_NAME", "SIGNATURE", "ID") VALUES (:a1, :a2, :a3, :a4)  [["first_name", "Peter"], ["last_name", "Doe"], ["signature", "cGV0ZXInc3NpZ25hdHVyZQ==\n"], ["id", 1]]
```

### System configuration
**Rails version**:
Same as test code in master.

**Oracle enhanced adapter version**:
Master branch.

**Ruby version**:
2.7.3p183

**Oracle Database version**:
19.3.0

akostadinov:
I see that the `write_lob` hooks are only registered on UPDATE. Is there any special reason for that? What is the reason not to be registered on SAVE as well?

https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced/lob.rb#L14-L15

PR #2268, opened 2022-02-28
workaround deserialize on save - rails/rails#44317

akostadinov:
This is more of a hack but I'm not sure of a better way to fix. Provided that we can use `#write_lobs` for such inserts, then I don't know why we should serialize and insert with the insert statement thus need the `#serialize`? Related to #2226

I feel there should be a better fix but I'm confused with the whole idea around lob writing.


yahonda:
Related to https://github.com/rails/rails/issues/44317

akostadinov:
Basically if we always use `#write_lobs` perhaps we don't need to have a `#serialize` method and thus no fixing would be needed.

And if we believe that `#write_lobs` is necessary is some use cases, then this or another workaround would be needed. Or Rails to stop updating the attribute value unnecessarily.

akostadinov:
bump

PR #2280, opened 2022-03-15
Keep GitHub Actions updated with Dependabot

yahonda:
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot


yahonda:
We do not push `Gemfile.lock` to this repository, we can disable `package_manager: "ruby:bundler"` .

PR #2282, opened 2022-03-18
Require MFA to release

koic:
## Summary

Follow up https://github.com/rails/rails/commit/9195b7f and https://github.com/rails/rails/commit/1fde031.

This product can tell gem users that it has been released using MFA.
For example, users can see it with `REQUIRES MFA ACCOUNTS` on the Rails gem page:
https://rubygems.org/gems/rails

It's up to @yahonda as the lead maintainer to decide whether or not to take this.

## Other Information

There are other items that can be set in the `s.metadata`, but that is not the purpose of this PR.


yahonda:
I have already configured MFA to push gems and agree to set MFA by default.

PR #2283, opened 2022-03-30
Use default_sequence_name to identify sequence when rename a table

chriger:
I encountered a problem when renaming a table with a sequence that has a shortened table name in it to meet the 30 character limit.

Because errors when renaming a sequence are rescued with `nil` i recognized the error after executing the migration on my running test environment.

Example:
Table name: `HERE_IS_MY_LOOONG_TABLE_NAME`
Sequence name: `HERE_IS_MY_LOOONG_TABLE_NA_SEQ`

When executing a rename as follow:
```ruby
rename_table :here_is_my_looong_table_name :short_table_name
```

The adapter try to change a sequence named `HERE_IS_MY_LOOONG_TABLE_NAME_SEQ` instead of `HERE_IS_MY_LOOONG_TABLE_NA_SEQ`


yahonda:
Thank you for opening a pull request. Would you add some specs for this change?

PR #2289, opened 2022-04-28
CHAR should return nil for NULL values in column

sprileyMSTS:
lib/active_record/connection_adapters/oracle_enhanced/oci_connection.rb
- Return nil for NULL value in CHAR column similar to behavior in
  lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb


sprileyMSTS:
Let me know if it's possible to backport this fix into a patch release of the 6.0 and 6.1 versions of this adapter. If not I'll have to release something internally at my company to address this issue until we can upgrade to Rails 7.

yahonda:
Thanks for opening a pull request. Would you add some specs? 
Unfortunately, ActiveRecord Oracle enhanced adapter 6.1.z and 6.0.z are not actively maintained so no back port will be done.

Issue #2256, opened 2022-01-21
OracleEnhancedAdapter.emulate_booleans doesn't work


areman:
On load of OracleEnhancedAdapter the TYPE_MAP will be initialized. But that makes it impossible to set OracleEnhancedAdapter.emulate_booleans = false. To set it to false you need to require the ruby source file, and OracleEnhancedAdapter.emulate_booleans = false is always set after the TYPE_MAP is nitialized.

andynu:
Here is a monkey patching initializer workaround. https://gist.github.com/andynu/9522d16ab85fd4ec2c2c31903fbe249d

andynu:
Would the maintainers be interested in a PR along the lines of the workaround? Converting the TYPE_MAP constant into a class attribute? I'd be happy to provide one, but wasn't sure if that is the direction you'd like to go since it would involve re-parsing the mappings later in the load process. I'm not sure if there is a better approach to more lazily set the type_map so that the configs can be set before the first time it is initialized.

yahonda:
I am happy to review pull requests with tests. https://github.com/rsim/oracle-enhanced/pull/2199 aims to reduce amount of memory in Rails 7.0. So converting the TYPE_MAP into constant may introduce the same memory usage as Rails 6.1 or older.

yahonda:
Closed via #2301

PR #2199, opened 2021-07-29
Define adapter type maps statically when possible

yahonda:
This pull request supports  `Define adapter type maps statically when possible` https://github.com/rails/rails/pull/42773 in Active Record Oracle enhanced adapter. 

Also, to make CI green this pull request needs these additional commits.
* Use Oracle Instant Client 21.4
* Rename `_quote` to `quote`
* Rename `_type_cast` to `type_cast`

PR #2296, opened 2022-06-15
Lazy load type mappings

andynu:
This appears to fix issue #2256. I believe it should have the same memory implications as the intention in #2199. The type maps are loaded once when they are first called. Only one instance of the mapping is ever created.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -691,6 +691,12 @@ module ActiveRecord
       end
 
       class << self
+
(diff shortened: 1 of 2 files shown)


andynu:
This fixes the new test case. And works in my applications. Unfortunately, it also triggers some other test failures. Sadly, without being able to reproduce your test environment locally (https://github.com/yahonda/rails-dev-box/issues/43), I do not have the means to pursue the other test failures.

yahonda:
Would you create a pull request to master branch first? I'll backport it to release70 branch if necessary.

andynu:
Sure. #2300

PR #2298, opened 2022-07-04
Support use_foreign_keys? via rails/rails#45301

yahonda:
This pull request supports use_foreign_keys? via rails/rails#45301.


yahonda:
Bug report templates failures will be fixed after this pull request is merged.

PR #2299, opened 2022-07-04
Revert "Configure `continue-on-error: true` to run all CI"

yahonda:
Reverts rsim/oracle-enhanced#2130

As a side effect, setting `continue-on-error: true` marks the status "Success" even if the Jobs fails .  

Refer to https://github.com/rsim/oracle-enhanced/actions/runs/2606963371


yahonda:
The background is I was not aware that https://github.com/rsim/oracle-enhanced/pull/2298 needed to make CI green.

PR #2300, opened 2022-07-04
Lazy load type mappings

andynu:
(This time against the master branch)

This appears to fix issue https://github.com/rsim/oracle-enhanced/issues/2256. I believe it should have the same memory implications as the intention in https://github.com/rsim/oracle-enhanced/pull/2199. The type maps are loaded once when they are first called. Only one instance of the mapping is ever created.


yahonda:
Would you also change your branch based on the current master branch? It looks this pull request is trying to merge something from release7-0 branch with your commits to the master branch.

andynu:
Cleaner version without the no-op release70 commits mixed in here: #2301

PR #2301, opened 2022-07-05
Lazy load type mappings

andynu:
This appears to fix issue #2256. I believe it should have the same memory implications as the intention in #2199. The type maps are loaded once when they are first called. Only one instance of the mapping is ever created.

(clean patch against the master branch, nothing from the release70 here)


andynu:
I'd be delighted if this PR were merged. If there is anything I can do to help that along I'd be happy to do it.

Meanwhile @JRDVN Here is a monkey patching initializer workaround. https://gist.github.com/andynu/9522d16ab85fd4ec2c2c31903fbe249d

PR #2308, opened 2022-09-29
Add dbconsole support to adapter

gmcgibbon:
Related to https://github.com/rails/rails/pull/46093

Adds dbconsole method to be used when invoking bin/rails dbconsole.

Moves dbconsole logic related to Oracle in Rails to the adapter. This connection code and corresponding test was in the dbconsole command in railties.

cc @paarthmadan


lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
```diff
@@ -64,19 +64,22 @@
 
 module ActiveRecord
   module ConnectionHandling # :nodoc:
-    # Establishes a connection to the database that's used by all Active Record objects.
-    def oracle_enhanced_connection(config) # :nodoc:
+    def oracle_enhanced_connection_class # :nodoc:
```
gmcgibbon, line 67: Looks like we're probably going to rename this to `<adapter_name>_adapter_class`, I'll make that change ASAP.

yahonda:
I'll take care of it as #2499

PR #2319, opened 2023-03-22
Use Ubuntu 22.04

yahonda:
https://github.blog/changelog/2022-08-09-github-actions-ubuntu-22-04-is-now-generally-available-on-github-hosted-runners/


yahonda:
While the specs are failing, I have confirmed CI is running using the Ubuntu 22.04 image.

PR #2320, opened 2023-03-22
Use Oracle Instant Client 21.9.0.0.0

yahonda:
https://www.oracle.com/jp/database/technologies/instant-client/linux-x86-64-downloads.html


yahonda:
Confirmed Oracle Instant Client 21.9.0.0.0 is installed.

PR #2321, opened 2023-03-22
CI against Ruby 3.2

yahonda:
This pull request adds CI against Ruby 3.2 at GitHub Actions.


yahonda:
While specs are failing, at least CI is running with Ruby 3.2 as expected https://github.com/rsim/oracle-enhanced/actions/runs/4486077689/jobs/7888231057?pr=2321

PR #2333, opened 2023-05-10
Use Ubuntu 22.04 (Jammy Jellyfish) at Travis CI

yahonda:
https://docs.travis-ci.com/user/reference/jammy/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os: linux
-dist: focal
+dist: jammy
 cache: bundler
 
 env:


yahonda:
Ruby 3.0 and Ruby 2.7 installation are failing that is expected.

https://app.travis-ci.com/github/rsim/oracle-enhanced/jobs/601850143#L369
>     OpenSSL is not available. Install OpenSSL and rebuild Ruby (preferred) or use non-HTTPS sources


https://bugs.ruby-lang.org/issues/18658
> We decided to not backport openssl-3.x gem into ruby_2_7 and ruby_3_0 branches. Because it affects with OpenSSL 1.0.x.

yahonda:
Closing because it still needs to run Ruby 2.7. Will revisit once we do not need to run Ruby 3.0 or lower.

PR #2360, opened 2024-02-21
Enhance workaround for WeakMap with jruby >= 9.4.6.0

iaddict:
Since jruby 9.4.6.0 the missing methods on ObjectSpace::WeakMap have been implemented.

See: https://github.com/jruby/jruby/pull/6683

Without this fix the gem does not work on jruby 9.4.6.0.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -13,10 +13,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: Set up Ruby 3.2
+      - name: Set up Ruby 3.3
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 3.2
+          ruby-version: 3.3
       - name: Install required package
         run: |
           sudo apt-get install alien
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,7 @@ jobs:
         # - jruby,
         # - jruby-head
         ruby: [
+          '3.3',
           '3.2',
           '3.1',
           '3.0',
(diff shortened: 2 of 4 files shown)


laumacirule:
While pending on a new version, for the activerecord-oracle_enhanced-adapter 6.1.6 that we are currently using, we implemented a temporary workaround to rescue the NameError in `lib/active_record/connection_adapters/oracle_enhanced_adapter.rb`
```ruby
begin
  load File.expand_path("active_record/connection_adapters/oracle_enhanced_adapter.rb", $:.grep(/oracle_enhanced/).first)
rescue NameError => e
  raise e unless e.message =~ /undefined field 'map'/
end
```

rammpeter:
Another workaround to patch the gem tech time after fresh install like in CI pipelines:
add this snippet as last step in RAILS_ROOT/config/boot.rb
```
# Workaround to avoid the error: undefined field 'map' for class 'Java::OrgJruby::RubyObjectSpace::WeakMap'
# with oracle-enhanced-adapter 6.1.6 and JRuby 9.4.6.0 and following versions
# Prevent execution of the code block in the if statement by comparing RUBY_ENGINE with a not existing value
# This patches the gem each time it is installed again
# Peter Ramm, 2024-06-18

file_path = File.join(Dir.pwd, 'vendor', 'bundle', 'jruby', '3.1.0', 'gems', 'activerecord-oracle_enhanced-adapter-6.1.6', 'lib', 'active_record', 'connection_adapters', 'oracle_enhanced_adapter.rb')
content = File.read(file_path)
new_content = content.gsub("if RUBY_ENGINE == \"jruby\"", "if RUBY_ENGINE == \"xjruby\"")
File.open(file_path, 'w') { |file| file.write(new_content) }
```

Issue #2329, opened 2023-05-08
Support https://github.com/rails/rails/pull/44591


yahonda:
Oracle enhanced adapter needs to support https://github.com/rails/rails/pull/44591 which has eight commits.  https://github.com/rails/rails/commit/deec3004d8d85443dc4f3f5fd22ab86b10adb58b triggers unit test failures.

### Steps to reproduce
```ruby
bundle exec rspec ./spec/active_record/connection_adapters/emulation/oracle_adapter_spec.rb:12
```

### Expected behavior
It should pass.

### Actual behavior
```ruby
$ bundle exec rspec ./spec/active_record/connection_adapters/emulation/oracle_adapter_spec.rb:12
Fetching https://github.com/kubo/ruby-oci8.git
Fetching https://github.com/rsim/ruby-plsql.git
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Using rake 13.0.6
Using concurrent-ruby 1.2.2
Using connection_pool 2.4.0
Using minitest 5.18.0
Using ast 2.4.2
Using bundler 2.4.10
Using io-console 0.6.0
Using diff-lcs 1.5.0
Using json 2.6.3
Using parallel 1.23.0
Using stringio 3.0.6
Using rack 3.0.7
Using rainbow 3.1.1
Using regexp_parser 2.8.0
Using rexml 3.2.5
Using rspec-support 3.12.0
Using ruby-progressbar 1.13.0
Using i18n 1.13.0
Using tzinfo 2.0.6
Using reline 0.3.3
Using parser 3.2.2.1
Using psych 5.1.0
Using rspec-core 3.12.2
Using rspec-expectations 3.12.3
Using rspec-mocks 3.12.5
Using ruby-plsql 0.8.0 from https://github.com/rsim/ruby-plsql.git (at master@89a68f9)
Using activesupport 7.1.0.alpha from https://github.com/rails/rails.git (at /home/yahonda/src/github.com/rails/rails@50f8b85)
Using rspec 3.12.0
Using ruby-oci8 2.2.12 from https://github.com/kubo/ruby-oci8.git (at master@29290ee)
Using irb 1.6.4
Using rdoc 6.5.0
Using rubocop-ast 1.28.1
Using debug 1.7.2
Using unicode-display_width 2.4.2
Using rubocop 1.50.2
Using rubocop-capybara 2.18.0
Using rubocop-factory_bot 2.22.0
Using rubocop-performance 1.17.1
[...]

PR #2363, opened 2024-03-12
Bump Rails version to https://github.com/rails/rails/commit/0e92677

yahonda:
This commit bumps Rails version to https://github.com/rails/rails/commit/0e92677

Here is the difference between the current master branch and this commit. https://github.com/rails/rails/compare/8551e64e2411811f26d210601abdba6e13d8798c...0e92677


yahonda:
This is a part of #2329

Issue #2310, opened 2022-10-31
Support https://github.com/rails/rails/pull/46273


yahonda:
### Steps to reproduce
This is a placeholder to support https://github.com/rails/rails/pull/46273

### Expected behavior
Oracle enhanced adapter `execute` method should support `allow_retry` argument.

### Actual behavior
This change has not been implemented yet.

### System configuration
**Rails version**: master

**Oracle enhanced adapter version**: master

**Ruby version**:

**Oracle Database version**:

PR #2366, opened 2024-03-19
Add allow_retry kwarg to execute

andynu:
This extends the existing class level auto_retry functionality to optionally allow an allow_retry boolean kwarg to allow retry of specific queries.

This PR preserves the requirement that `autocommit?` be enabled which is the safer choice.

```
  def with_retry(allow_retry: false) # :nodoc:
    should_retry = (allow_retry || self.class.auto_retry?) && autocommit?
```

See issue #2310


andynu:
Dup of #2374

PR #2371, opened 2024-03-25
Support Rails 7.1 ColumnDefinition validations  (5)

andynu:
Support Rails 7.1 ColumnDefinition validations.

Rails commit e6da3ebd6c6 [1] introduces ColumnDefinition validations for
the keywords used in database commands. This commit adds the additional
keywords supported by this gem.

[1] https://github.com/rails/rails/commit/e6da3ebd6c65af23d134a9e01145f26600912008

Stacked on #2370

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :development do
   gem "rubocop-rails", require: false
   gem "rubocop-rspec", require: false
 
-  gem "activerecord",   github: "rails/rails", ref: "436277da88507f9aae0874e62f3e61a8546b9683"
+  gem "activerecord",   github: "rails/rails", ref: "e6da3ebd6c65af23d134a9e01145f26600912008"
   gem "ruby-plsql", github: "rsim/ruby-plsql", branch: "master"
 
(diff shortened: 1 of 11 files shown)


davinlagerroos:
Something to consider: you can also define an adapter-specific `valid_column_definition_options` that calls `super` and adds the db-specific extras, e.g. https://github.com/umn-asr/oracle-enhanced/commit/211e3d164ac94fbdd4d165c4e7e16a96922b581e. This is closer to the pattern that the reference implementations for mysql and postgres in activerecord use. For example:
https://github.com/rails/rails/pull/46178/files#diff-af97fb79600884bcb2fd1f3db7237c228f732d9a65f217949ede18525ec9a07d
and
https://github.com/rails/rails/pull/46178/files#diff-af97fb79600884bcb2fd1f3db7237c228f732d9a65f217949ede18525ec9a07d

andynu:
Thanks @davinlagerroos, your suggestion of overriding TableDefinition#valid_column_definition_options is a better approach. I've updated this PR accordingly.

PR #2377, opened 2024-03-26
Switch to Arel::Table.name instead of removed alias table_name.

andynu:
Switch to Arel::Table.name instead of removed alias table_name.

Rails removed the table_name alias to Arel::Table.name
See https://github.com/rails/rails/commit/1d98bc563a8a7faf26238eaaa54a3257a95d644d

Stacked on #2375


andynu:
Will open an alternate.

PR #2381, opened 2024-03-31
Make insert return an array, and include the id_value if super returns nil (12)

### LINKED ISSUE #2380: "Assign auto populated columns on Active Record object creation", breaks the returning of id on insert
### Overview
The rails commit [1] changes how values are returned from insert statements. The worthy goal is to use the RETURNING statement in postgres to set all the database generated attribute values as a direct result of the insert statement. Oracle supports the RETURNING statement too and may ultimately benefit from this change! As part of this change, the ActiveRecord::Persistence#_create_record now looks for a set of `returning_columns` instead of assuming that the primary key is being returned.

The challenge with this in the oracle adapter is the way this `returning_columns` list is built via 
ActiveRecord::ConnectionAdapters::Column#auto_incremented_by_db? which defaults to a literal false. (or slightly higher up the callstack `connection.return_value_after_insert?(c)` where c is a Column)

The oracle enhanced adapter does not appear to have knowledge about if a column is a primary key or not on the Column record itself (at least not on that record, or within the sql_type_metadata). It is unclear to me how best to identify which columns are the primary key, and thus should be included in the `returning_columns` collection.

[1] https://github.com/rails/rails/commit/3421e892afa532a254e54379ac2ce9bef138cf3f

### Steps to reproduce
You can see the barest implementation of the parameters for the rails change, and the problem in the tests here: 
https://github.com/andynu/oracle-enhanced/commit/336d04b588f62f97814142528f8b99d44e32ecf4

### Expected behavior
insert should return the id of the new record

### Actual behavior
insert returns nil

### System configuration
**Rails version**: not quite 7.1.0 - 3421e892afa532a254e54379ac2ce9bef138cf3f

**Oracle enhanced adapter version**: a little ways off master, 4707dfa45fcaa73df5399626f752204165016f7b

**Ruby version**: 3.2.2

**Oracle Database version**: 23.0.0

andynu:
Make insert return an array, and include the id_value if super returns nil

ActiveRecord::ConnectionAdapters::OracleEnhanced::DatabaseStatements#insert
The super here is not returning the id_value if it is available (as it used
to). Furthermore ActiveRecord::Persistence#_create_record expects an
array of values.

I don't think this is the right change. In fact I think that the original abstract
ActiveRecord::ConnectionAdapters::DatabaseStatements#insert
may not be handling id_value correctly. The method appears to sometimes return
an array, and other times return a singular id_value.

However, I do think this highlights the crux of supporting rails 7.1 in this library as with this little tweak, all the tests passing against rails 7-1-stable.

This fixes #2380

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :development do
   gem "rubocop-rails", require: false
   gem "rubocop-rspec", require: false
 
-  gem "activerecord",   github: "rails/rails", ref: "436277da88507f9aae0874e62f3e61a8546b9683"
+  gem "activerecord",   github: "rails/rails", ref: "e6da3ebd6c65af23d134a9e01145f26600912008"
   gem "ruby-plsql", github: "rsim/ruby-plsql", branch: "master"
 
   platforms :ruby do
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :development do
   gem "rubocop-rails", require: false
   gem "rubocop-rspec", require: false
 
-  gem "activerecord",   github: "rails/rails", ref: "59f2b06493bed2c036f6b527e743198416ee1237"
+  gem "activerecord",   github: "rails/rails", ref: "3a1662f71434351f7d08687ae2ae6d5f565d2c99"
   gem "ruby-plsql", github: "rsim/ruby-plsql", branch: "master"
 
   platforms :ruby do
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :development do
   gem "rubocop-rails", require: false
   gem "rubocop-rspec", require: false
 
-  gem "activerecord",   github: "rails/rails", ref: "3a1662f71434351f7d08687ae2ae6d5f565d2c99"
+  gem "activerecord",   github: "rails/rails", ref: "d12f1a2f84128085ee0bdbbbe28df9ec27a9bc64"
   gem "ruby-plsql", github: "rsim/ruby-plsql", branch: "master"
 
   platforms :ruby do
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :development do
   gem "rubocop-rails", require: false
   gem "rubocop-rspec", require: false
 
-  gem "activerecord",   github: "rails/rails", ref: "e6da3ebd6c65af23d134a9e01145f26600912008"
+  gem "activerecord",   github: "rails/rails", ref: "8f780c58ea2a5af17d61ac5858b33ead2c7aa494"
   gem "ruby-plsql", github: "rsim/ruby-plsql", branch: "master"
 
   platforms :ruby do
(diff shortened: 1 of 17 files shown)


mattalat:
Hi @andynu, thank you for your PRs. I had a few questions.

1. What is the intention of the 12 separate MRs? Do they provide different changes or are they all stacked in this  twelfth one?

2. There is a code comment that mentions needing to figure out if a column is a PK/auto-incremented. In https://github.com/rsim/oracle-enhanced/issues/2380 you mention that this may have been fixed upstream. Is it still an ongoing concern?

3. Regarding your notes on the structure of the abstract adapter’s `#insert` method, do you have a scenario to which you could direct me that showcases this behavior? The method’s documentation mentions it could return an Array of values if specific conditions are met, but I wasn’t sure if they are in Oracle. And, if it does seem like a bug on the adapter’s end, do you know if anyone has identified it in the Rails repo?

Thanks for your time

andynu:
Hi @mattalat 

1. The purpose of the stack of PRs is to show passing tests along the way. That does mean that each one in the stack contains the changes from the previous, and that this last one contains all the changes. As I mentioned in the rails 7.1 upgrade issue, I'd be happy to re-form these PRs as independent PRs instead of a stack, but with those you would not be able to see the passing tests because unresolved issues from other PRs would still be breaking the tests.

There are two ways to go about reviewing these changes. First we can go in order, that first PR is a minimal change on top of master, and if/when that one were merged in then the second PR in the stack would become simplified to just another small change. However, I acknowledge that some of these PRs may not be the implementation that the maintainers would like to use, in which case I would look at their fixes to rebase subsequent PRs on their changes. If you'd like the unstacked diff for any of these you need to do a comparison with the previous PR e.g. https://github.com/andynu/oracle-enhanced/compare/stack-09-internal_exec_query...andynu:oracle-enhanced:stack-12-sql-statement-args (sorry that the branch names are not nicely numerically sequential, the bisecting to produce this stack of PRs was a challenging journey with many false starts and backpeddling).

2.  So the issue in #2380 has been fixed in rails 7.2, and backported to 7.1. However, in this PR in DatabaseStatements#insert you can see `Array(super || id_value)` which I think is really working around a bug in rails, where super should be providing an array itself when an id_value is provided, but it doesn't it just returns whatever id_value is passed in, even though its api has changed to always expect an array to come out of it. I do think a better workaround involves adding support for OracleEnhanced::Column.auto_incremented_by_db?. I would put a pin in this one until the other fixes are sorted out, and then it'd be great to have a second set of eyes verify this fix /wrt issue #2381 and the upstream changes.

3. You can definitely see the failure. If you check out c0439811da4bc7f8cb310db78bb69c9be9dd3c69 (That's one commit back in this PR), then you'll see a pile of test failures that illustrate the issue. I tried to have the meat of each of these PRs start with a version bump, and then a fix for the test failures that showed up. My stack of PRs here was explicitly chasing test failures of tests that already exist in this library as I bumped the rails version closer and closer to 7.1 stable.

Issue #2357, opened 2024-01-15
Please privide Rails 7.1 compatibility


drgcms:
### Steps to reproduce
Database connection 

### Expected behavior
Success

### Actual behavior
[oracle-enhanced (deb214decc37) lib/active_record/connection_adapters/oracle_enhanced_adapter.rb:699:in `get_database_version'](http://razvoj:3340/register-finance#)
[activerecord (7.1.2) lib/active_record/connection_adapters/schema_cache.rb:374:in `database_version'](http://razvoj:3340/register-finance#)
[activerecord (7.1.2) lib/active_record/connection_adapters/schema_cache.rb:70:in `database_version'](http://razvoj:3340/register-finance#)
[activerecord (7.1.2) lib/active_record/connection_adapters/schema_cache.rb:200:in `database_version'](http://razvoj:3340/register-finance#)
[activerecord (7.1.2) lib/active_record/connection_adapters/abstract_adapter.rb:871:in `database_version'](http://razvoj:3340/register-finance#)
[oracle-enhanced (deb214decc37) lib/active_record/connection_adapters/oracle_enhanced_adapter.rb:307:in `supports_fetch_first_n_rows_and_offset?'](http://razvoj:3340/register-finance#)
[oracle-enhanced (deb214decc37) lib/active_record/connection_adapters/oracle_enhanced_adapter.rb:271:in `arel_visitor'](http://razvoj:3340/register-finance#)
[activerecord (7.1.2) lib/active_record/connection_adapters/abstract_adapter.rb:159:in `initialize'](http://razvoj:3340/register-finance#)
[oracle-enhanced (deb214decc37) lib/active_record/connection_adapters/oracle_enhanced_adapter.rb:250:in `initialize'](http://razvoj:3340/register-finance#)
[oracle-enhanced (deb214decc37) lib/active_record/connection_adapters/oracle_enhanced_adapter.rb:76:in `new'](http://razvoj:3340/register-finance#)
[oracle-enhanced (deb214decc37) lib/active_record/connection_adapters/oracle_enhanced_adapter.rb:76:in `oracle_enhanced_connection'](http://razvoj:3340/register-finance#)
[activerecord (7.1.2) lib/active_record/connection_adapters/abstract/connection_pool.rb:673:in `public_send'](http://razvoj:3340/register-finance#)
[...]

PR #2384, opened 2024-07-05
Rails 7.x Compatibility

hss-mateus:
Related to #2357 

This PR just applies all the patches provided by @andynu since now, plus some temporary workarounds to get it working with a simple model.

I have tested this branch on a Rails 7.2.0.beta2 application, and it seems to be working for simple queries/models.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :development do
   gem "rubocop-rails", require: false
   gem "rubocop-rspec", require: false
 
-  gem "activerecord",   github: "rails/rails", ref: "436277da88507f9aae0874e62f3e61a8546b9683"
+  gem "activerecord",   github: "rails/rails", ref: "e6da3ebd6c65af23d134a9e01145f26600912008"
   gem "ruby-plsql", github: "rsim/ruby-plsql", branch: "master"
 
   platforms :ruby do
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :development do
   gem "rubocop-rails", require: false
   gem "rubocop-rspec", require: false
 
-  gem "activerecord",   github: "rails/rails", ref: "59f2b06493bed2c036f6b527e743198416ee1237"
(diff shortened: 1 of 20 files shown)


yahonda (commented):
Please ignore my previous update.

yahonda:
Thanks for your contribution. Let me merge this pull request and address some TODO/known issues.

akostadinov:
Hi, this PR forces the old visitor (`Oracle` instead of `Oracle12`) and this is a breaking change. Why shouldn't we force the new visitor instead? And also no documented way to select the desired visitor. Did nobody else get unexpected results with this version?

cc @yahonda

andynu:
@akostadinov Rails 7.x is out of main support. And security support for 7.2 ends in 3 months. I highly encourage you to upgrade.

https://endoflife.date/rails

akostadinov:
@andynu this change remains active in master:
https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L353-L360

It seems to have been done as a temporary hack and remained. Or do you intentionally want to keep this hardcoded to the Oracle 11 visitor?

P.S. Very recently my changes were refused, I assume because they don't apply to Oracle 11 which is EOL since longer. So I respectfully don't appreciate the EOL sentiment in the post. (let me use the opportunity to point out that Oracle11 carries such a technical depth and obviously creates real confusing issues, that I don't understand why should we actively keep it supported as a higher tier than mainstream versions)

andynu:
Fair point if the behavior you describe is still on master. Commenting on a 2 year old pull request is confusing from my end. If you would like to submit an issue with a specific demonstration of the "unexpected results" I'd be happy to continue the discussion there.

akostadinov:
FYI this is supposedly fixed with https://github.com/rsim/oracle-enhanced/pull/2573, thanks to yahonda, who beat me to it and fixed it in a better way to what I started to do.

P.S. commenting on the source of the issue to potentially notify the original authors and for a reference whenever somebody hits it. Like I find it useful for myself when I see additional context in old issues/PRs/threads. So I assume might be useful to others. Regards.

PR #2385, opened 2024-08-13
Use Oracle Instant Client 23.5.0.0.0

yahonda:
This pull request updates the Oracle Instant Client version to 23.5.0.0.0 to use same Oracle database / client version.
https://www.oracle.com/database/technologies/instant-client/linux-x86-64-downloads.html


yahonda:
I'll address CI failures later.

PR #2413, opened 2024-10-07
Skip the patch for moderm JRuby vers

ryudoawaru:
By the comment [here](https://github.com/jruby/jruby/issues/6267#issuecomment-2395573502), O believe it's not necessary to apply the patch for moderm JRuby versions.


rammpeter:
It would be very helpful for JRuby users to complete this simple improvement by merging the PR.
Just now it raises additional trouble while making oracle_enhanced tests functional with JRuby 10.0.2.0

PR #2415, opened 2024-10-16
Use IS [NOT] NULL instead of DBMS_LOB.COMPARE for nil CLOB/BLOB queries

### LINKED ISSUE #2410: 7.1 update: strange serialize behavior
Currently, I try to update my app from rails 7.0 to 7.1.

One thing that I recognized it the behavior of searching for DB entries where `serialize` is used for a column to store a `Hash`

Let's say, I have the model:

```ruby
class SomeModel < ApplicationRecord
  serialize :something, type: Hash, coder: YAML
end
```

I would like to get all entries where `something` is empty/nil/null/...

Using Rails 7.0:

```ruby
SomeModel.where('something is NULL').count
# --> "SELECT \"SOME_MODELS\".* FROM \"SOME_MODELS\" WHERE (something is NULL)"
# result is 100
```

```ruby
SomeModel.where(something: nil).count
# --> "SELECT \"SOME_MODELS\".* FROM \"SOME_MODELS\" WHERE \"SOME_MODELS\".\"SOMETHING\" IS NULL"
# result is 100
```

```ruby
SomeModel.where(something: {}).count
# --> "SELECT \"SOME_MODELS\".* FROM \"SOME_MODELS\" WHERE \"SOME_MODELS\".\"SOMETHING\" IS NULL"
# result is 100
```

I always get the same result.

When I do the same with Rails 7.1

```ruby
SomeModel.where('something is NULL').count
# --> "SELECT \"SOME_MODELS\".* FROM \"SOME_MODELS\" WHERE (something is NULL)"
# result is 100
```

```ruby
SomeModel.where(something: nil).count
# -->"SELECT \"SOME_MODELS\".* FROM \"SOME_MODELS\" WHERE DBMS_LOB.COMPARE(\"SOME_MODELS\".\"SOMETHING\", NULL) = 0"
# result is 0
```

```ruby
SomeModel.where(something: {}).count
# -->"SELECT \"SOME_MODELS\".* FROM \"SOME_MODELS\" WHERE DBMS_LOB.COMPARE(\"SOME_MODELS\".\"SOMETHING\", NULL) = 0"
# result is 0
```

I only get the correct result for `SomeModel.where('something is NULL').count`. The other queries do not work.

Is this a new Rails 7.1 behavior, or is this related to the Oracle adapter? I would not be a fan of switching from `something: nil` to `'something is NULL'` because I would need an additional `where` if I would further filter the entries, and rubocop would complain anyway (`Use `where(something: nil)` instead of manually constructing SQL. (convention:Rails/WhereEquals)`).

### System configuration
**Rails version**: `7.1.4`

**Oracle enhanced adapter version**: `7.1.0`

**Ruby version**: `3.3.4`

**Oracle Database version**: `19.22 (19c)`

brianthoman:
Closes #2410

### DIFF
--- a/lib/arel/visitors/oracle_common.rb
+++ b/lib/arel/visitors/oracle_common.rb
@@ -9,7 +9,9 @@ module Arel # :nodoc: all
         # Fixes ORA-00932: inconsistent datatypes: expected - got CLOB
         def visit_Arel_Nodes_Equality(o, collector)
           left = o.left
+
           return super unless %i(text binary).include?(cached_column_for(left)&.type)
+          return super if o.right.nil?
 
           # https://docs.oracle.com/cd/B19306_01/appdev.102/b14258/d_lob.htm#i1016668
           # returns 0 when the comparison succeeds
--- a/lib/arel/visitors/oracle_common.rb
+++ b/lib/arel/visitors/oracle_common.rb
@@ -9,13 +9,14 @@ module Arel # :nodoc: all
         # Fixes ORA-00932: inconsistent datatypes: expected - got CLOB
         def visit_Arel_Nodes_Equality(o, collector)
           left = o.left
+          right = o.right
 
+          return super if right.nil?
           return super unless %i(text binary).include?(cached_column_for(left)&.type)
-          return super if o.right.nil?
 
           # https://docs.oracle.com/cd/B19306_01/appdev.102/b14258/d_lob.htm#i1016668
           # returns 0 when the comparison succeeds
-          comparator = Arel::Nodes::NamedFunction.new("DBMS_LOB.COMPARE", [left, o.right])
+          comparator = Arel::Nodes::NamedFunction.new("DBMS_LOB.COMPARE", [left, right])
           collector = visit comparator, collector
           collector << " = 0"
           collector
--- a/spec/active_record/oracle_enhanced/type/binary_spec.rb
+++ b/spec/active_record/oracle_enhanced/type/binary_spec.rb
@@ -26,6 +26,7 @@ describe "OracleEnhancedAdapter handling of BLOB columns" do
   after(:all) do
     @conn.drop_table :test_employees, if_exists: true
     Object.send(:remove_const, "TestEmployee")
+    Object.send(:remove_const, "TestSerializedHashEmployee")
   end
 
   after(:each) do
--- a/spec/active_record/oracle_enhanced/type/binary_spec.rb
+++ b/spec/active_record/oracle_enhanced/type/binary_spec.rb
@@ -137,21 +137,21 @@ describe "OracleEnhancedAdapter handling of BLOB columns" do
   it "should find serialized NULL BLOB data when queried with nil" do
     TestSerializedEmployee.delete_all
     TestSerializedEmployee.create!(binary_data: nil)
-    TestSerializedEmployee.create!(binary_data: { data: 'some data' })
+    TestSerializedEmployee.create!(binary_data: { data: "some data" })
     expect(TestSerializedEmployee.where(binary_data: nil)).to have_attributes(count: 1)
   end
 
   it "should find serialized Hash NULL BLOB data when queried with nil" do
     TestSerializedHashEmployee.delete_all
     TestSerializedHashEmployee.create!(binary_data: nil)
-    TestSerializedHashEmployee.create!(binary_data: { data: 'some data' })
+    TestSerializedHashEmployee.create!(binary_data: { data: "some data" })
     expect(TestSerializedHashEmployee.where(binary_data: nil)).to have_attributes(count: 1)
   end
 
   it "should find serialized Hash NULL BLOB data when queried with {}" do
     TestSerializedHashEmployee.delete_all
     TestSerializedHashEmployee.create!(binary_data: nil)
-    TestSerializedHashEmployee.create!(binary_data: { data: 'some data' })
+    TestSerializedHashEmployee.create!(binary_data: { data: "some data" })
     expect(TestSerializedHashEmployee.where(binary_data: {})).to have_attributes(count: 1)
   end
 end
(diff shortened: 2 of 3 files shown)


lib/arel/visitors/oracle_common.rb
```diff
@@ -9,7 +9,9 @@ module OracleCommon
         # Fixes ORA-00932: inconsistent datatypes: expected - got CLOB
         def visit_Arel_Nodes_Equality(o, collector)
           left = o.left
+
           return super unless %i(text binary).include?(cached_column_for(left)&.type)
+          return super if o.right.nil?
```
akostadinov, line 14: Just a nitpick, maybe create a variable `right` and check it for being `nil` earlier because it seems a more lightweight check than the other one. But it's fine as it is too.

spec/active_record/oracle_enhanced/type/binary_spec.rb
```diff
@@ -116,4 +120,32 @@ class ::TestEmployee < ActiveRecord::Base
...
+    TestEmployee.create!(binary_data: nil)
+    TestEmployee.create!(binary_data: @binary_data)
+    expect(TestEmployee.where(binary_data: nil)).to have_attributes(count: 1)
+  end
+
+  it "should find serialized NULL BLOB data when queried with nil" do
+    TestEmployee.delete_all
+    TestEmployee.create!(binary_data: nil)
+    TestEmployee.create!(binary_data: @binary_data)
+    expect(TestEmployee.where(binary_data: nil)).to have_attributes(count: 1)
```
akostadinov, line 135: Maybe it's too late here but I don't see a difference with the previous test.

spec/active_record/oracle_enhanced/type/binary_spec.rb
```diff
@@ -116,4 +120,32 @@ class ::TestEmployee < ActiveRecord::Base
...
+  end
+
+  it "should find serialized NULL BLOB data when queried with nil" do
```
akostadinov, line 138: here the test name appears to be the same as the previous one

spec/active_record/oracle_enhanced/type/text_spec.rb
```diff
@@ -31,6 +31,10 @@ class ::TestEmployee < ActiveRecord::Base; end
     class ::Test2Employee < ActiveRecord::Base
       serialize :comments
     end
+    class ::TestSerializedHashEmployee < ActiveRecord::Base
```
akostadinov, line 34: maybe this class name should be `Test2SerializedHashEmployee` because in `spec/active_record/oracle_enhanced/type/binary_spec.rb` there is already a class `TestSerializedHashEmployee`.

brianthoman: I thought those classes were scoped to the `describe` block, so there shouldn't be any conflict. E.g., there's already a `TestEmployee` class defined in both `binary_spec.rb` and `text_spec.rb`.

Happy to make this change though, if that's the safer route.

brianthoman:
Thank you for the review @akostadinov! I made some changes based on your comments.

spec/active_record/oracle_enhanced/type/text_spec.rb
akostadinov: If the describe block scopes these anyhow, I don't see how it would. More likely to me, Ruby just reopens the class and changes it as it goes which seems messy to me.

spec/active_record/oracle_enhanced/type/text_spec.rb
```diff
@@ -241,4 +245,25 @@ class ::TestSerializeEmployee < ActiveRecord::Base
...
+    TestEmployee.create!(comments: nil)
+    TestEmployee.create!(comments: @char_data)
+    expect(TestEmployee.where(comments: nil)).to have_attributes(count: 1)
+  end
+
+  it "should find serialized NULL CLOB data when queried with nil" do
+    Test2Employee.delete_all
+    Test2Employee.create!(comments: nil)
+    Test2Employee.create!(comments: { some: 'text' })
+    expect(Test2Employee.where(comments: nil)).to have_attributes(count: 1)
```
akostadinov, line 260: As far as I understand, this test should be using the `TestSerializedHashEmployee` or `TestSerializedHashEmployee2` class. Because it is named `... serialized ...`. Let me know if I didn't get your idea right.

spec/active_record/oracle_enhanced/type/text_spec.rb
brianthoman: Yeah, you're right. I think I conflated this with my understanding of [helper methods in RSpec](https://rspec.info/features/3-13/rspec-core/helper-methods/arbitrary-methods/).

I also overlooked that the existing specs seem to be [dealing with this using `remove_const`](https://github.com/brianthoman/oracle-enhanced/blob/5f07e232028cb187b4301cde8155f58dae4f2a12/spec/active_record/oracle_enhanced/type/text_spec.rb#L52-L56) after all the specs have run. 

I've added a `remove_const` to clean up the new test class I added. Do you think that addresses the issue? If not, I'll finally give in and change the name 😅. Maybe to something like `TextSpecSerializedHashEmployee` so we don't have to keep track of what Test Employee number we're up to across files.

spec/active_record/oracle_enhanced/type/text_spec.rb
brianthoman: I think `TestEmployee2` or `TestSerializeEmployee` should work for this test case, as both have `serialize :comments`, but I switched it to `TestSerializeEmployee` to keep things matched up a little better.

akostadinov (approved):
Thank you for fixing this and for the updates! To me everything seems alright now.

akostadinov:
@brianthoman , would you rebase and squash the commits (it seems like maintainers want squashed commits in general).

PR #2418, opened 2024-11-13
Fix rename_table for compatibility migrations

michael-smith-nz:
👋
This PR fixes the `ArgumentError: wrong number of arguments (given 3, expected 2)` issue when calling `rename_table` with the migration version 7.0 and below.

It looks like ActiveRecord is expecting `rename_table` to include `options` as a parameter when calling `super` [here](https://github.com/rails/rails/blob/7-1-stable/activerecord/lib/active_record/migration/compatibility.rb#L118)


yahonda:
Related to https://github.com/rails/rails/pull/50837

Issue #2411, opened 2024-10-04
Support for Rails 7.2


dwarner:
I know 7.1.0 just got released with support for Rails 7.1.x (thanks), but is there an ETA for Rails 7.2 support?

https://github.com/rsim/oracle-enhanced/blob/v7.1.0/activerecord-oracle_enhanced-adapter.gemspec
`s.add_runtime_dependency("activerecord", ["~> 7.1.0"])`

I know Rails 7.2 ActiveRecord has a lot of "Remove deprecated" . . .
https://guides.rubyonrails.org/7_2_release_notes.html#active-record

numbcoder:
+1

matthewkrom:
+1

rjschof:
+1. Any updates on this or any sort of plans to maintain this gem?

andynu:
I started, and got stuck. https://github.com/rsim/oracle-enhanced/pull/2422
Anyone willing to give my very limited progress a look, I'd appreciate additional perspectives and/or suggestions for diagnosing.

andynu:
I've made further progress and cut a new PR for consideration. Not all the way there, but a significant step closer. There are now only four remaining test failures locally.  #2424

dwarner:
@andynu Thanks for your hard work!

We've pulled your 7.2 branch into one of our projects and have NOT run into a problem.

We'll be trying it on other projects coming up.

Do you have an ETA on an official release?

andynu:
Hi Dennis, I'm glad the patch has been working well for you, and me too!
Alas, I am just a random user; I have no access to make releases.

akostadinov:
Whoever tested this, maybe you add an approval so hopefully author attention is captured.

joshRpowell:
> [@andynu](https://github.com/andynu) Thanks for your hard work!
> 
> We've pulled your 7.2 branch into one of our projects and have NOT run into a problem.
> 
> We'll be trying it on other projects coming up.
> 
> Do you have an ETA on an official release?

+1 branch has been running in app ~6mo now. No issues. 
Thanks for all your work @andynu 🥳 
(unfortunately, not able to approve)

dwarner:
@andynu I will make a comment on the PR. 

Have you gotten any responses/feedback from the maintainers?

dwarner:
@rsim Are you still maintaining this project?

smarks-1-ND:
[...]

PR #2424, opened 2025-01-31
WIP Rails 7 2 support

andynu:
This makes significant progress towards supporting Rails 7.2

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :development do
   gem "rubocop-rails", require: false
   gem "rubocop-rspec", require: false
 
-  gem "activerecord",   github: "rails/rails", branch: "7-1-stable"
+  gem "activerecord",   github: "rails/rails", branch: "7-2-stable"
   gem "ruby-plsql", github: "rsim/ruby-plsql", branch: "master"
 
   platforms :ruby do
--- a/activerecord-oracle_enhanced-adapter.gemspec
+++ b/activerecord-oracle_enhanced-adapter.gemspec
@@ -26,7 +26,7 @@ This adapter is superset of original ActiveRecord Oracle adapter.
     "rubygems_mfa_required" => "true"
   }
 
-  s.add_runtime_dependency("activerecord", ["~> 7.1.0"])
+  s.add_runtime_dependency("activerecord", ["~> 7.2.0"])
   s.add_runtime_dependency("ruby-plsql", [">= 0.6.0"])
   if /java/.match?(RUBY_PLATFORM)
     s.platform = Gem::Platform.new("java")
--- a/lib/active_record/connection_adapters/oracle_enhanced/context_index.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/context_index.rb
@@ -330,8 +330,23 @@ module ActiveRecord
             # Add context index condition.
             def contains(column, query, options = {})
               score_label = options[:label].to_i || 1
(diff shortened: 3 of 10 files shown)


andynu:
- Emulate OracleAdapter isn't working because the loading process changed significantly. [oracle_enhanced_connection](https://github.com/rsim/oracle-enhanced/blob/792ef9a64ba9eafdae7cee1e7deaec910b6b6331/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L68) is never called.
- The fix here for Model.contains context index searching is a work around for that specific case, but I'm worried that there is a more general conversion from activerecord bind variables '?' to compatible oracle bind variables :N that is not well handled.
- The timezone tests do not reliably run for me, perhaps oracle client version related.

dlagerro:
I've been looking at contains errors for a fork we have at work. The `?` binds issue is bigger than `contains`, it affects anytime a user tries to write a query with a `where` with a `?`, like `Book.where("title = ?", params[:title])`. The cause seems to be that the oracle visitor does not define a bind block, so it falls back to the [arel use of `?`](https://github.com/rails/rails/blob/v7.2.2.1/activerecord/lib/arel/visitors/to_sql.rb#L751-L754). 

Adding
```
      BIND_BLOCK = proc { |i| ":a#{i}" }
      private_constant :BIND_BLOCK

      def bind_block; BIND_BLOCK; end
```
to `lib/arel/visitors/oracle.rb` fixes the issue for me.

The bind_block is essentially what `visit_ActiveModel_Attribute` and `visit_Arel_Nodes_BindParam` are doing, so with the bind block set, those methods are no longer needed.

andynu:
@dlagerro Thank you!!!! I knew I was papering over a broader issue in the hopes of being able to resolve other issues here first. I had spent more time on the broader issue but not made the breakthrough you appear to have made. I will look at your proposed change and revert the AREL conversion of the `contains` method. Thank you so much.

andynu:
@dlagerro's suggestion worked great for the bind parameters issue, and was similar to changes found in the postgres adapter.

I've also changed how you invoke the emulate_oracle_adapter feature. You used to have a database config named `emulate_oracle_adapter` and there was a handler that would dynamically require the OracleAdapter subclass of the OracleEnhancedAdapter class. That dynamic inclusion has moved into the framework itself where it is now easy to register different classes to match with the different `adapter` keywords. So I've replaced the feature here to use that functionality.
Usually, people will want and use `adapter: oracle_enhanced` which maps to the OracleEnhancedAdapter class. And if you want to get an emulated OracleAdapter class, all you need to do is set your `adapter: oracle` and you'll get it, no special database config option needed anymore.

Further, new versions of activerecord require ruby 3.1+ so those older rubies failing is expected. The only reason I did not update the test matrix is because of a comment in the config about jruby.

andynu:
I have removed ruby 2.7 and 3.0 from the test matrix. They are not compatible with Rails 7.2

dwarner:
Thanks for the work, @andynu!

We have this up and running in one of our projects with no issues. 

We will be testing in more projects in the coming weeks.

@joshRpowell says he has the "branch has been running in app ~6mo now. No issues"
https://github.com/rsim/oracle-enhanced/issues/2411#issuecomment-2884025571

Is the failing rubocop build the last thing holding up a release?

Anybody know which maintainer is taking lead on the approval of the PR?

yahonda:
Thank you!

PR #2428, opened 2025-03-12
Log load error - master branch specs passes

jlahtinen:
Wanted to propose a way to log schema:load error ddl in some way.

Specs did not pass so I tried to make them pass. I had to make some changes that I am quite unsure are they good changes or not ... But specs passes now. This depends on one change to ruby-plsql gem. I make pull request to that also next.


jlahtinen:
https://github.com/rsim/ruby-plsql/pull/235

jlahtinen:
I was running oracle in container tag is free:23.6.0.0
With jruby 9.4.12.0

PR #2467, opened 2025-07-09
Fix #2429: Replaced slow UNION query with optimized all_objects query

### LINKED ISSUE #2429: Slow query, can it be replaced?
https://github.com/rsim/oracle-enhanced/blob/d5b3daf8bdac7578055467329bcb6600114ea47a/lib/active_record/connection_adapters/oracle_enhanced/connection.rb#L37-L56

The above query is hitting us hard in Production. I'm seeing ~2s responses for this query. Our dbas have suggested that it could be replaced with the following:
```
select owner,object_name from all_objects where owner=:owner object_name=:obj_name and object_type in ('TABLE','VIEW','SYNONYM');
```

I don't know enough about the inner workings of Oracle to know whether this is a useful improvement and whether it would work over all the supported versions of Oracle for this Gem. Can someone with a bit more knowledge provide me with some insight here?

LohithaNadimpalli:
## Summary:
Replaced the slow UNION-based SQL query in `connection.rb` with a single query using `all_objects` to improve performance, as discussed in #2429.

## Details:
- Reduced 4 UNION ALL queries into one optimized query using `all_objects`.
- Suggested by Oracle DBAs due to 2-second delay in production.
- I'm a beginner contributor and wasn’t able to run tests locally.
  Kindly review and let me know if further changes are needed.

Closes #2429

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
@@ -34,25 +34,11 @@ module ActiveRecord
               table_owner, table_name = default_owner, real_name
             end
             sql = <<~SQL.squish
-              SELECT owner, table_name, 'TABLE' name_type
-              FROM all_tables
-              WHERE owner = :table_owner
-                AND table_name = :table_name
-              UNION ALL
-              SELECT owner, view_name table_name, 'VIEW' name_type
-              FROM all_views
-              WHERE owner = :table_owner
-                AND view_name = :table_name
-              UNION ALL
-              SELECT table_owner, table_name, 'SYNONYM' name_type
-              FROM all_synonyms
-              WHERE owner = :table_owner
-                AND synonym_name = :table_name
-              UNION ALL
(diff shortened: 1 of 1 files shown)


matthewtusker:
We've done some more research into this and found a `DBMS_UTILITY.NAME_RESOLVE` function in Oracle which resolves names, and supports db links. It's much faster than the existing SQL, and faster than the `all_objects` query. I think you should look into that if you're looking to make a contribution to Oracle Enhanced.

yahonda:
Thanks for the contribution and the idea to use `all_objects`. I have opened #2521 which builds on this approach with the following fixes:

- Alias columns (`object_name table_name`, `object_type name_type`) to match the existing result handling code
- Fix bind parameter count (3 instead of 7)
- Restore public synonym lookup via `OR (owner = 'PUBLIC' AND object_name = :real_name)`
- Resolve synonym targets via a second `all_synonyms` query (since `all_objects` does not return the synonym target)
- Add `ORDER BY` to preserve the original lookup priority
- Add a regression test covering table, view, private synonym and public synonym resolution

PR #2469, opened 2025-07-11
select compatible AR version while developing

akostadinov:
Without this change, running tests locally selects the latest available activerecord version locally. This might be incorrect and confusing as hell.

This change ensures that a supported version of activerecord is selected by bundler. Also it is very standard to have this in Gemfiles.

I hope this doesn't break CI. So I'm submiting anyway and hope it works.


akostadinov:
I see that you want to test with latest ruby :thinking: 
Perhaps I better update the README. I will close for now.

Issue #2470, opened 2025-07-14
Method " select_statement?" missing in OracleEnhanced::JDBCConnection::Cursor


rammpeter:
In class OracleEnhanced::DatabaseStatements line 57 a method  `
if cursor.select_statement?` is called.

This method `select_statement?`exists for OracleEnhanced::OCIConnection::Cursor but is missed for OracleEnhanced::JDBCConnection::Cursor.

Thus using ActiveRecord with JRuby 10.0.0.1 fails with 
```
>> Dual.select "SELECT SYSDATE FROM DUAL"
An error occurred when inspecting the object: #<ActiveRecord::StatementInvalid: NoMethodError: undefined method 'select_statement?' for an instance of ActiveRecord::ConnectionAdapters::OracleEnhanced::JDBCConnection::Cursor>
An error occurred when running Kernel#inspect: #<ActiveRecord::StatementInvalid: NoMethodError: undefined method 'select_statement?' for an instance of ActiveRecord::ConnectionAdapters::OracleEnhanced::JDBCConnection::Cursor>
/Users/pramm/.rubies/jruby-10.0.0.1/lib/ruby/gems/shared/bundler/gems/oracle-enhanced-695537daabd5/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb:57:in 'block in raw_execute'
/Users/pramm/.rubies/jruby-10.0.0.1/lib/ruby/gems/shared/gems/activesupport-8.0.2/lib/active_support/notifications/instrumenter.rb:58:in 'instrument'
/Users/pramm/.rubies/jruby-10.0.0.1/lib/ruby/gems/shared/bundler/gems/oracle-enhanced-695537daabd5/lib/active_record/connection_adapters/oracle_enhanced/dbms_output.rb:35:in 'log'
/Users/pramm/.rubies/jruby-10.0.0.1/lib/ruby/gems/shared/bundler/gems/oracle-enhanced-695537daabd5/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb:33:in 'block in raw_execute'
/Users/pramm/.rubies/jruby-10.0.0.1/lib/ruby/gems/shared/gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract_adapter.rb:1015:in 'block in with_raw_connection'
[...]

PR #2473, opened 2025-08-25
Added missing method Cursor.select_statement?

rammpeter:
This method was missed for the JDBC implementation of class Cursor

Fixes issue #2470

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
@@ -389,6 +389,13 @@ module ActiveRecord
             @raw_statement.getUpdateCount
           end
 
+          # Is the current statement a SELECT statement?
+          def select_statement?
+            # Only simple SELECT and WITH statements are considered SELECT statements.
+            # because no other valid ojdbc method found to check it.
+            @raw_statement.get_original_sql.strip.match?(/\A\s*(SELECT|WITH)/i)
+          end
+
           def fetch(options = {})
             if @raw_result_set.next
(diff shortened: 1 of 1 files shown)


jlahtinen:
Seems to be working.

I noticed there are fixes in release80 branch but there is no release done that contains those fixes. Is it clear to someone how/when they are released? Will the release contain this fix? What about the currently unnecessary # Workaround for https://github.com/jruby/jruby/issues/6267 etc

yahonda:
For awareness: `JDBCConnection::Cursor#select_statement?` has already been added on `master` in [`0929019b`](https://github.com/rsim/oracle-enhanced/commit/0929019bab91ef6ed1c847a7f1503f80b7bd0e78) with a simpler implementation:

```ruby
def select_statement?
  !@raw_result_set.nil?
end
```

On `master` this works because an earlier refactor ([`4d7e9fe8`](https://github.com/rsim/oracle-enhanced/commit/4d7e9fe8dbc9bae8a2a9bdceca657ff88b07dd09) — "Use createStatement for DDL / PL/SQL in JDBC prepare") changed `Cursor#exec` to assign `@raw_result_set` only when `@raw_statement.execute` returns `true` (i.e. the first result is a `ResultSet`), which is JDBC's authoritative SELECT signal. On `release80`, `exec` still calls `executeQuery` unconditionally, so the same one-liner wouldn't distinguish SELECT from DML there — which is why this PR's regex-on-SQL workaround is needed on that branch.

So `master` doesn't need this change. Whether to merge this PR as-is on `release80`, or to backport `4d7e9fe8` + `0929019b` instead, is a separate release80-only decision.

PR #2498, opened 2026-04-08
Refactor OracleEnhanced::Connection responsibilities

yahonda:
## Summary
- Split the base `OracleEnhanced::Connection` into a thin dispatcher and a new `DatabaseDescription` module so each piece has a single responsibility.
- `Connection.create` now delegates to a registered `connection_class` instead of branching on `ORACLE_ENHANCED_CONNECTION` (the constant is kept because `database_statements.rb` and the spec suite still read it).
- `describe` / `_select_one` / `_select_value` / `_oracle_downcase` move to `OracleEnhanced::DatabaseDescription`, included by both `OCIConnection` and `JDBCConnection`. This removes the `def describe(name); super; end` shims that only existed to expose the base class's private method.

## Test plan
- [ ] Run `bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/connection_spec.rb` on MRI
- [ ] Run the same spec on JRuby (JDBC path)
- [ ] Run schema-related specs that exercise `_connection.describe(table_name)` via `schema_statements.rb` / `oracle_enhanced_adapter.rb`

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/connection.rb
@@ -5,113 +5,15 @@ module ActiveRecord
     # interface independent methods
     module OracleEnhanced
       class Connection # :nodoc:
+        class << self
+          attr_accessor :connection_class
+        end
+
         def self.create(config)
-          case ORACLE_ENHANCED_CONNECTION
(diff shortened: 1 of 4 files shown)


yahonda:
Closing this in favor of #2545

PR #2513, opened 2026-04-10
Use DBMS_METADATA for structure dump

yahonda:
## Summary

Adds a second `structure_dump` implementation built on Oracle's `DBMS_METADATA.GET_DDL` / `GET_DEPENDENT_DDL` alongside the existing data-dictionary (`ALL_*`) one. The new path lives in a separate file (`lib/active_record/connection_adapters/oracle_enhanced/structure_dump/dbms_metadata.rb`) — the existing `structure_dump.rb` is **untouched**, so selecting `:data_dictionary` retains exact master behavior.

## Selection

`OracleEnhancedAdapter.structure_dump_method` accepts:

- **`:auto`** (default) — DBMS_METADATA on Oracle 12.1+, otherwise data-dictionary. The 12.1 floor is a project-policy version gate (the IDENTITY / EDITIONABLE / modern `SET_TRANSFORM_PARAM` era), exposed as `use_dbms_metadata_dump?`. `DBMS_METADATA.GET_DDL` itself works on Oracle 9i+; this is a project decision, not a strict capability check.
- **`:dbms_metadata`** — force DBMS_METADATA. Raises `ArgumentError` on pre-12.1 (mirrors PR #2576's `identifier_max_length: :long` fail-fast policy).
- **`:data_dictionary`** — force the implementation that assembles DDL from the `ALL_*` static data dictionary views in Ruby. Remains explicitly selectable on 12.1+ as well.

## Implementation shape

The DBMS_METADATA path walks `ALL_OBJECTS` per object_type rather than hand-rolling per-kind queries. This naturally picks up `MATERIALIZED VIEW`, `TYPE`, `TRIGGER`, `SYNONYM` and friends without separate code paths and minimizes future maintenance when Oracle adds new object kinds. FK constraints (`REF_CONSTRAINT`) are emitted in a single trailing block; cross-table dependency order is left to the user (or to a fresh-schema reload) rather than tracked here.

Two modules in the same file split responsibilities so the dispatcher logic doesn't bleed into the DBMS_METADATA implementation:

- `Dispatcher` — `prepend`-ed; routes the three entry points (`structure_dump`, `structure_dump_db_stored_code`, `structure_dump_synonyms`) to either DBMS_METADATA or `super` (data-dictionary).
- `DbmsMetadata` — `include`-d; the DBMS_METADATA implementation itself, with no dispatch logic.

## Output

Output shape is informed by `mysqldump --no-data` / `pg_dump --schema-only`:

- `STORAGE` / `TABLESPACE` / `SEGMENT_ATTRIBUTES` / `EMIT_SCHEMA` suppressed via `DBMS_METADATA.SET_TRANSFORM_PARAM` so the dump is portable across installations and schemas. Oracle honors these for `TABLE` / `INDEX` / `SEQUENCE`.
- **`MATERIALIZED VIEW` is an Oracle-side exception**: even with object_type-specific overrides Oracle does not honour the suppression on MV, so MV DDL retains physical attributes (`PCTFREE`, `TABLESPACE`, `SEGMENT CREATION`, etc.) as Oracle emits them. Since MVs aren't creatable through Rails' standard migration helpers anyway, no post-processing pass is added — Oracle's output is used as-is.
- `REF_CONSTRAINTS` emitted as separate `ALTER TABLE … ADD CONSTRAINT` statements after all tables.
- Inline constraints via Oracle's `CONSTRAINTS=TRUE` default; UNIQUE backing indexes that DBMS_METADATA already inlines into the table DDL are filtered out so they aren't emitted twice.
- `STATEMENT_TOKEN` separation preserved (Oracle's `SQLTERMINATOR=FALSE` default) so `execute_structure_dump` splits as before.
- `primary_key_trigger:` (#2615) row triggers are picked up via the `TRIGGER` object_type pass — addresses [#2513 (comment)](https://github.com/rsim/oracle-enhanced/pull/2513#issuecomment-4331026696).
- COMMENT ON TABLE / COMMENT ON COLUMN queried directly because `GET_DEPENDENT_DDL("COMMENT", ...)` is unreliable.
- `structure_dump_db_stored_code` selects `'PROCEDURE', 'PACKAGE', 'FUNCTION', 'TYPE'` only — `GET_DDL("PACKAGE", ...)` already returns spec + body, so selecting `'PACKAGE BODY'` would emit the body twice.
- Only `SET_TRANSFORM_PARAM` calls whose value differs from Oracle's default are issued (`STORAGE` / `TABLESPACE` / `SEGMENT_ATTRIBUTES` / `EMIT_SCHEMA` / `REF_CONSTRAINTS`).

## Tests

A new `dbms_metadata_structure_dump_spec.rb` exercises the new path with **structural** assertions only (`CREATE TABLE` / `CREATE INDEX` / `ALTER TABLE … ADD CONSTRAINT` / `COMMENT ON …` / no `STORAGE` / `TABLESPACE` / `PCTFREE` noise) — exact byte-level DDL is intentionally not part of the contract, mirroring how `pg_dump` / `mysqldump` are tested upstream. Regression coverage added for the unique-index and PACKAGE-BODY duplicate-output bugs found during development. A schema-option spec verifies that `:schema` connection-time switching is honored (the dump walks `SYS_CONTEXT('userenv', 'current_schema')`, not the connecting user).

The existing `structure_dump_spec.rb` pins `structure_dump_method = :data_dictionary` in `before(:all)` and restores it in `after(:all)`, so the data-dictionary backend's exact-DDL assertions continue to run.

## Test plan

- [x] `bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/structure_dump_spec.rb` — green (pinned to `:data_dictionary`)
- [x] `bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/dbms_metadata_structure_dump_spec.rb` — green
- [x] `bundle exec rspec` full suite — 747 examples, 0 failures, 12 pending

Closes #2513.


yahonda:
Heads-up: #2615 (currently open) restores trigger-based primary key generation as the `primary_key_trigger:` opt-in. If #2615 lands, `db:structure:dump` should also pick up the BEFORE INSERT row trigger emitted by `create_pk_trigger` so that `db:structure:load` can recreate the table + sequence + trigger faithfully.

`DBMS_METADATA.GET_DEPENDENT_DDL('TRIGGER', table_name, schema)` returns the trigger DDL as a CLOB, which fits the `GET_DDL`-based approach this PR is taking. Worth adding when both PRs settle, either to this PR (if it stays open until #2615 merges) or as a follow-up.

The Ruby-flavored `db/schema.rb` round-trip is already handled inside #2615 via `SchemaDumper`; this note is purely about the `db/structure.sql` path.

PR #2514, opened 2026-04-11
Configure ORA_TZFILE to match Oracle 11g server in CI

yahonda:
## Summary

- Both 11g CI workflows (`test_11g.yml`, `test_11g_ojdbc11.yml`) used `DATABASE_SERVER_AND_CLIENT_VERSION_DO_NOT_MATCH=true` to mask ORA-01805 errors caused by Oracle 11g XE shipping DSTv14 while Instant Client 21.15 embeds v35. Adopt the approach from rsim/ruby-plsql#239: copy the v14 data files out of the running `gvenzl/oracle-xe:11` container into the Instant Client's `oracore/zoneinfo` directory and point the client at them with `ORA_TZFILE`. `ORA_TZFILE` is set to just the basename (`timezlrg_14.dat`) so Instant Client resolves it relative to `$ORACLE_HOME/oracore/zoneinfo/`; the absolute-path form produces `ORA-01804` in sqlplus. This lets the timestamp-with-timezone specs run against 11g instead of being skipped.
- Drop `DATABASE_SERVER_AND_CLIENT_VERSION_DO_NOT_MATCH` from both 11g workflows and remove the corresponding skip in `spec/active_record/oracle_enhanced/type/timestamp_spec.rb`. Also remove the stale `ORACLE_HOME == "/usr/lib/oracle/21/client64"` fallback there — current workflows use `/opt/oracle/instantclient_21_15`, so it never matched.
- Rewire the two JDBC network-encryption skips in `connection_spec.rb` onto `DATABASE_VERSION == "11.2.0.2"`, which is the honest condition — 11g XE genuinely lacks native network encryption regardless of client version.

## Test plan

- [x] `test_11g` workflow: new `Configure ORA_TZFILE to match Oracle 11g server` step succeeds and the timestamp-with-timezone examples run (previously skipped) without ORA-01805.
- [x] `test_11g_ojdbc11` workflow: same as above.
- [x] The two JDBC network-encryption examples in `connection_spec.rb` remain skipped in the 11g jobs with reason "Oracle 11g XE does not support native network encryption".
- [x] `test.yml` (Oracle 23c) is unaffected — it never set the removed env var and does not touch `ORA_TZFILE`.

### DIFF
--- a/.github/workflows/test_11g.yml
+++ b/.github/workflows/test_11g.yml
@@ -36,7 +36,6 @@ jobs:
       DATABASE_HOST: localhost
       DATABASE_PORT: 1521
       DATABASE_VERSION: 11.2.0.2
-      DATABASE_SERVER_AND_CLIENT_VERSION_DO_NOT_MATCH: true
 
     services:
       oracle:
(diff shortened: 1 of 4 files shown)


yahonda:
Backport to release81 branch 0877c09d7188e4110ff6aac7858c622a9323d933

PR #2521, opened 2026-04-13
Replace slow UNION ALL query with optimized all_objects query

yahonda:
## Summary
- Replace the 4-way `UNION ALL` query across `all_tables`, `all_views`, and `all_synonyms` inside `OracleEnhanced::SchemaStatements#resolve_data_source_name` with a single `all_objects` query for the common TABLE/VIEW lookup path.
- Resolve synonym targets via a second `all_synonyms` query only when a synonym is found.
- `ORDER BY DECODE(...)` preserves the legacy lookup priority (owner table, owner view, owner synonym, public synonym) in the single-pass query.
- Add a regression spec covering table, view, materialized view, private synonym, and public synonym resolution for one underlying table.
- Grant `CREATE/DROP PUBLIC SYNONYM` to test users so the new spec can run from a fresh setup.

## Motivation
The original `UNION ALL` query causes ~2s delays in production (see #2429). This replaces it with a single `all_objects` query, which is significantly faster for the common TABLE/VIEW path. Based on the approach suggested in #2467.

## Note on rebase
Rebased onto current `master` (past #2545 "Move describe to schema_statements as resolve_data_source_name"). The optimization now lives in `OracleEnhanced::SchemaStatements#resolve_data_source_name` rather than the former `Connection#describe`. Upstream's iterative loop and `Set`-based synonym cycle guard are preserved; the catalog lookup still goes through `select_one(..., "SCHEMA", ...)` so it participates in instrumentation and the query cache.

## Test plan
- [x] Local `bundle exec rspec .../connection_spec.rb` — 77 examples, 0 failures, 1 pending (JDBC-only).
- [x] `resolve_data_source_name` block — 14 examples, 0 failures (includes the new combined regression).
- [x] `bundle exec rubocop` clean on both Ruby files.
- [x] Copilot CLI review (upstream/master..HEAD) — no actionable findings.
- [x] Verified materialized views are still found (backing table appears as `TABLE` in `all_objects`); combined regression spec now covers this case explicitly.

Co-Authored-By: N.lohitha <lohithavarma36@gmail.com>
Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>
Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -758,30 +758,37 @@ module ActiveRecord
               binds = [
                 bind_string("table_owner", owner),
                 bind_string("table_name", identifier),
-                bind_string("table_owner", owner),
-                bind_string("table_name", identifier),
-                bind_string("table_owner", owner),
-                bind_string("table_name", identifier),
                 bind_string("real_name", real_name),
               ]
+              # Single-pass lookup against all_objects, ordered so the first row
+              # is the one the legacy 4-way UNION ALL (all_tables, all_views,
+              # owner synonym, public synonym) would have returned: prefer a
+              # match in the caller's schema over PUBLIC, and within a schema
+              # prefer TABLE over VIEW over SYNONYM.
               result = select_one(<<~SQL.squish, "SCHEMA", binds)
-                SELECT owner, table_name, 'TABLE' name_type
-                FROM all_tables WHERE owner = :table_owner AND table_name = :table_name
-                UNION ALL
-                SELECT owner, view_name table_name, 'VIEW' name_type
-                FROM all_views WHERE owner = :table_owner AND view_name = :table_name
-                UNION ALL
-                SELECT table_owner, table_name, 'SYNONYM' name_type
-                FROM all_synonyms WHERE owner = :table_owner AND synonym_name = :table_name
-                UNION ALL
-                SELECT table_owner, table_name, 'SYNONYM' name_type
-                FROM all_synonyms WHERE owner = 'PUBLIC' AND synonym_name = :real_name
+                SELECT owner, object_name table_name, object_type name_type
+                FROM all_objects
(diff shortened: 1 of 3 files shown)


matthewtusker:
Can `DBMS_UTILITY.NAME_RESOLVE` be used here instead? We're using it in Production (via a Monkey Patch) and it's working perfectly.

yahonda:
Opened a PoC https://github.com/rsim/oracle-enhanced/pull/2531

yahonda:
## Why this approach over the alternatives

Two alternatives were benchmarked head-to-head against this PR (23c and 11g, 900-object fixture, full numbers in #2566):

**#2566 (DBMS_UTILITY.NAME_RESOLVE)** — 2–3.5x faster than this PR on 23c, up to ~600x faster on the 11g synonym path. Rejected here because it replaces the SQL catalog query with a PL/SQL call (anonymous block on OCI / `CallableStatement` on JDBC), which is a style break from the adapter's uniform `ALL_*` dictionary access. The synonym-heavy advantage isn't representative of typical Rails on Oracle workloads. Parked as draft; revisit only if a concrete synonym-heavy workload is reported where the measured speedup clearly outweighs the added surface.

**#2560 (USER_* fallback for 11g)** — adds a `same_schema_as_user?` helper plus three hand-picked `ALL_*` → `USER_*` swaps, explicitly scheduled for removal when 11.2 support drops. This PR subsumes the `resolve_data_source_name` bucket (~86% of #2560's measured 11g wall-clock savings) without per-query branching. #2560's `synonyms` / `foreign_keys` wins (~50s on the 11g suite) would remain orthogonal, but the per-query dual code paths weren't judged worth carrying given most of the 11g pain is subsumed here. #2560 closed.

In short: this PR takes the middle path — large 11g win for the common table/view path, readable single-query SQL, and no per-version branching; the remaining synonym-heavy 11g headroom sits in #2566 as insurance if it's ever shown to matter in practice.

PR #2522, opened 2026-04-13
Remove File.exists? workaround for ruby-plsql

yahonda:
## Summary
- Remove the `File.exists?` shim from `spec/spec_helper.rb` that was added as a workaround for ruby-plsql using the removed `File.exists?` method

## Context
`File.exists?` was deprecated in Ruby 2.1 and removed in Ruby 3.2. The `spec/spec_helper.rb` had a workaround aliasing `File.exist?` to `File.exists?` so that ruby-plsql's JDBC connection code could load under JRuby 10.x.

This workaround is no longer needed now that the upstream fix has been applied in rsim/ruby-plsql#267.

## Test plan
- [ ] Verify JRuby CI passes without the shim


yahonda:
ruby-plsql 0.9.9 has been released that includes https://github.com/rsim/ruby-plsql/pull/267

PR #2537, opened 2026-04-16
Test: #2521 after #2535

yahonda:
Verify PR #2521 (all_objects query) applies cleanly after PR #2535 (db link support) merges. Conflict in create_oracle_enhanced_users.sql resolved by combining both grant sets.


yahonda:
Closing the test has been done.

PR #2541, opened 2026-04-17
Log username on ORA-01017 authentication failure (OCI and JDBC)

yahonda:
## Summary

- When `OCI8.new` raises ORA-01017 (invalid username/password), print the username, connection string, and privilege to `$stderr` before re-raising
- Equivalent handling for JDBC: when `DriverManager.getConnection` / `ORACLE_DRIVER.connect` fails with `SQLException` error code 1017, print the username, JDBC URL, and privilege to `$stderr`

This makes CI failures caused by misconfigured credentials immediately diagnosable — instead of just seeing `ORA-01017`, you see which username and connection string were attempted.

## Test plan

- [ ] Verify OCI path: connect with wrong password → `ORA-01017: Authentication failed. username="..." connection_string="..." privilege=...` appears on stderr
- [ ] Verify JDBC path (JRuby): same behaviour on wrong-password connection attempt
- [ ] Verify correct credentials still connect successfully (no regression)

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
@@ -146,12 +146,19 @@ module ActiveRecord
             end
 
             begin
-              @raw_connection = java.sql.DriverManager.getConnection(url, properties)
-            rescue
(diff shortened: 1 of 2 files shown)


yahonda:
In general this change may expose credentials to log files. Closing this since it was originally for debug purpose.

PR #2560, opened 2026-04-20
Use USER_* dictionary views to speed up Oracle 11g

yahonda:
## Summary

Three targeted swaps from `ALL_*` to `USER_*` dictionary views, guarded by a new
`same_schema_as_user?` helper. They cut the local Oracle 11g XE suite from
~17:30 to ~11:48 (about 30% faster), with no observable regression on 23c
(still ~1:02 for the same 441 examples).

`ALL_*` dictionary views are substantially more expensive on Oracle 11g because
they carry cross-schema visibility logic that is not needed when the caller is
looking in its own schema. `USER_*` returns the same rows when
`SYS_CONTEXT('USERENV','CURRENT_SCHEMA')` equals the connected user, which is
the case for any connection not using the `schema:` connection option.

### Helper

`OracleEnhancedAdapter#same_schema_as_user?` returns `true` when the `schema:`
connection option is blank or points at the connected user. It memoizes the
answer since the `schema:` option only fires an `ALTER SESSION SET
CURRENT_SCHEMA` at connect time (see `oci_connection.rb:363` /
`jdbc_connection.rb:198`). `blank?` mirrors the connection setup, which treats
a blank `schema:` as "no override".

### Measured impact (11g XE, 441 examples)

| Query | Calls | Before | After | Savings |
|---|---|---|---|---|
| `SchemaStatements#synonyms` | 143 | 44.6 s | 6.5 s | −38 s |
| `SchemaStatements#resolve_data_source_name` UNION | 1399 | 668 s | 360 s | −308 s |
| `SchemaStatements#foreign_keys` | 175 | 13.8 s | 1.3 s | −12.5 s |
| **Wall-clock** (full suite) |  | 17:30 | 11:48 | **−5:13** |

23c shows no regression: 1:02 before and after, 0 failures.

### Swaps I tried and reverted

Not every `ALL_*` → `USER_*` swap helps on 11g. These were measured on the same
suite and reverted:

| Method | 11g result |
|---|---|
| `views` / `materialized_views` | neutral / mild regression |
| `table_exists?` | slightly slower (~7 ms/call worse) |
| `tables` (DECODE MINUS query) | neutral |
| `temporary_table?` | slight regression |
| `indexes` (big index metadata join) | **3× slower per call** (178 → 506 ms) |

For `indexes` in particular, `USER_INDEXES` picks a worse access plan on 11g
than the equivalent `ALL_INDEXES` WHERE `owner = SYS_CONTEXT(...)` — opposite
of the pattern in the committed three swaps.

### Out of scope

- Pre-existing limitation: the ALL_* `foreign_keys` query already required
  `r.owner = c.owner`, so cross-schema FK targets were never supported. The new
  USER_* branch preserves that behavior rather than introducing it.
- If user code calls `ALTER SESSION SET CURRENT_SCHEMA` outside the adapter's
  `:schema` option, `same_schema_as_user?` would stay `true` even though
  CURRENT_SCHEMA has moved. That is neither tested nor supported by the
  adapter today, and `default_tablespace` already assumes CURRENT_SCHEMA ==
  USER in its `user_users` query.
- Pre-existing OCI issue (not fixed here): `oci_connection.rb:33` does
  `@owner = config[:schema]; @owner ||= config[:username]`. Because `""` is
  truthy in Ruby, `schema: ""` leaves `@owner = ""` instead of falling back
  to the login user. JDBC handles this via `schema.blank?`; OCI does not.
  The new `USER_*` fast path uses the `USER` pseudo-column and so is not
  affected, but the remaining `ALL_*` branches still bind `owner = ""` for
  blank-schema OCI connections. Fixing that alignment is a separate PR.

### Intent when 11.2 support is eventually dropped

The three `same_schema_as_user?` branches exist to help Oracle 11.2 specifically
— the cost gap between `USER_*` and `ALL_*` on 23c is negligible. When 11.2
support is removed, these branches and the helper should be deleted, leaving
only the `ALL_*` form. That cleanup is deliberately left for the 11.2-removal
PR.

## Test plan

- [x] Full rspec suite on Oracle 23c FREEPDB1 — 0 failures, 441 examples
- [x] Full rspec suite on Oracle 11.2 XE — 0 failures, 441 examples, 9 pending
  (same pending set as before; 3 extra vs 23c are version-gated)
- [ ] CI `test.yml` (23c matrix)
- [ ] CI `test_11g.yml` (11g matrix — expected to drop from ~14-15 min to ~10-11 min per job)

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -75,10 +75,17 @@ module ActiveRecord
 
         # get synonyms for schema dump
         def synonyms
-          result = select_all(<<~SQL.squish, "SCHEMA")
-            SELECT synonym_name, table_owner, table_name
-            FROM all_synonyms where owner = SYS_CONTEXT('userenv', 'current_schema')
-          SQL
+          result = if same_schema_as_user?
+            select_all(<<~SQL.squish, "SCHEMA")
+              SELECT synonym_name, table_owner, table_name
+              FROM user_synonyms
+            SQL
+          else
+            select_all(<<~SQL.squish, "SCHEMA")
+              SELECT synonym_name, table_owner, table_name
+              FROM all_synonyms where owner = SYS_CONTEXT('userenv', 'current_schema')
+            SQL
+          end
 
           result.collect do |row|
              OracleEnhanced::SynonymDefinition.new(oracle_downcase(row["synonym_name"]),
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -544,26 +544,45 @@ module ActiveRecord
         def foreign_keys(table_name) # :nodoc:
           (_owner, desc_table_name) = resolve_data_source_name(table_name)
 
-          fk_info = select_all(<<~SQL.squish, "SCHEMA", [bind_string("desc_table_name", desc_table_name)])
-            SELECT r.table_name to_table
-                  ,rc.column_name references_column
-                  ,cc.column_name
-                  ,c.constraint_name name
-                  ,c.delete_rule
-              FROM all_constraints c, all_cons_columns cc,
-                   all_constraints r, all_cons_columns rc
-             WHERE c.owner = SYS_CONTEXT('userenv', 'current_schema')
-               AND c.table_name = :desc_table_name
-               AND c.constraint_type = 'R'
-               AND cc.owner = c.owner
-               AND cc.constraint_name = c.constraint_name
-               AND r.constraint_name = c.r_constraint_name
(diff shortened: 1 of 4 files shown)


yahonda:
Closing without merging. After benchmarking #2521 vs #2566 (DBMS_UTILITY.NAME_RESOLVE) on both 23c and 11g, the plan is:

- land #2521 first -- it closes most of the #2429 latency on 23c and buys ~2000x on the non-synonym path on 11g, with minimal blast radius;
- keep #2566 as a parked draft reference, to revisit only if a synonym-heavy workload is reported where #2521's secondary all_synonyms lookup becomes the bottleneck.

The ALL_* → USER_* fallback here adds a per-query same_schema_as_user? branch (already with documented 'tried and reverted' non-uniform cases, and explicitly scheduled for removal when 11.2 support drops). Given the alternatives above cover the same describe-path wins without the per-query conditional code path, this PR's complexity isn't worth carrying.

PR #2566, opened 2026-04-21
Resolve describe() via DBMS_UTILITY.NAME_RESOLVE

yahonda:
## Status: POC reference — implementation moving to a separate branch

**Update (2026-05-04):** revisited after #2513 (DBMS_METADATA structure_dump) merged. With `DBMS_METADATA.*` already in the adapter, using `DBMS_UTILITY.NAME_RESOLVE` is no longer a style break. The latest 23ai numbers (below) and the precedent set by #2513 tip the balance: this branch is being kept as a **POC reference** with the benchmark script committed, and the production implementation will land on a new topic branch.

**Decision rationale**:

1. `DBMS_METADATA` is now an accepted in-adapter dependency (PR #2513), so using another `DBMS_*` package no longer breaks the adapter's idiom.
2. Speedups on the synonym path are large enough to justify the OCI/JDBC OUT-bind code split — the same kind of split #2513 already carries.
3. The remaining concern from earlier ("readability of an anonymous PL/SQL block vs. a SQL query") is outweighed by the measured wins.

The implementation PR will:

- Re-implement `resolve_data_source_name` on top of `DBMS_UTILITY.NAME_RESOLVE` (carry the OCI/JDBC OUT-bind plumbing from this branch).
- Reuse the benchmark script (`script/benchmark_describe.rb`) that this PR committed.
- Update the circular-synonym specs from #2545 to expect `ORA-00980` (this branch already has the spec edits).

[Earlier 2026-04-21 note]: #2521 was merged and addressed the #2429 describe() latency on both 23c and the non-synonym 11g path. This PR remained parked pending a concrete report of a synonym-heavy workload; the latest benchmark below is that report.

## Latest benchmark (2026-05-04)

CRuby 4.0.3, **Oracle 23ai (gvenzl/oracle-free, FREEPDB1)**, 1000-object fixture (700 tables + 100 views + 100 private synonyms + 100 public synonyms), local container, no network hop. Avg ms per `resolve_data_source_name()` call (lower is better). Commits measured: master `9c25e2c7` (post-#2513 merge), this branch `99a2ca34`.

| case             | master (post-#2513) | **this PR** | speedup |
|------------------|--------------------:|------------:|--------:|
| tables           |               0.384 |   **0.300** |   1.3x  |
| views            |               0.364 |   **0.303** |   1.2x  |
| private synonyms |              13.031 |   **0.447** | **29x** |
| public synonyms  |               1.723 |   **0.454** |   3.8x  |
| all mixed        |               0.578 |   **0.265** |   2.2x  |

The **private synonym path on master degraded to ~13 ms/call** in this run (vs. 0.848 ms/call when this PR was first measured against an earlier master). The `all_synonyms` secondary lookup that #2521 left in place is the dominant cost on a 1000-object schema, exactly the failure mode this branch was designed to bypass. `DBMS_UTILITY.NAME_RESOLVE` chases synonyms inside the PL/SQL engine, sidestepping `ALL_SYNONYMS` entirely.

11g XE numbers are not re-measured here; the earlier `~600x` synonym-path speedup on `gvenzl/oracle-xe:11` from the original benchmark stands as the directional reference.

## Summary

Resolve table / view / synonym names via `DBMS_UTILITY.NAME_RESOLVE` instead of catalog queries against `ALL_TABLES` / `ALL_VIEWS` / `ALL_SYNONYMS`. Targets the ~2s describe() latency reported in #2429 and the synonym-path bottleneck #2521 leaves behind on synonym-heavy schemas.

NAME_RESOLVE chases private and public synonyms inside the PL/SQL engine and returns `[owner, object_name]` in one round trip, so the work does not depend on `ALL_*` dictionary-view cost — which is exactly where the slowness on 11g and on synonym-heavy 23ai schemas originates.

`DBMS_UTILITY.NAME_RESOLVE` is documented for Oracle 8i (1999), so it's available on every version this adapter realistically targets in 2026.

## Implementation notes (POC, kept here for reference)

Subtleties the spec suite surfaced:

1. `NAME_RESOLVE`'s `context` parameter matters. `context=1` is PL/SQL objects only and raises `ORA-04047` for tables/views. `context=0` covers tables, views, and synonyms (verified against Oracle 23c and 11g XE).

2. `NAME_RESOLVE` uppercases unquoted input, so case-preserving (quoted) identifiers like `"test_Mixed_Comments"` failed with `ORA-06564`. `normalize_name_for_name_resolve` wraps each dotted part in double quotes when the original identifier is not upcase-normalized by `valid_table_name?`, so `sys.test_Mixed` becomes `SYS."test_Mixed"` rather than the all-quoted `"sys"."test_Mixed"` (which would send Oracle hunting for a lowercase schema and miss SYS).

3. Circular synonyms surface as `ORA-00980` ("synonym translation is no longer valid") from NAME_RESOLVE, not the SQL layer's `ORA-01775` ("looping chain of synonyms"). `resolve_data_source_name` wraps that as an `OracleEnhanced::ConnectionException` — no Ruby-side stack overflow, clean error. The existing circular-synonym specs from #2545 are updated to match the new surfacing.

4. NAME_RESOLVE runs through an anonymous PL/SQL block (OCI) / `CallableStatement` (JDBC), bypassing `select_one`. The `sql.active_record` SCHEMA notification from #2545 is preserved via explicit `instrumenter.instrument`, so logging / subscribers continue to see the describe call.

OUT-bind plumbing lives on both the OCI8 path (anonymous PL/SQL block with named binds) and the JDBC path (`CallableStatement` + `registerOutParameter`) so this runs on MRI and JRuby alike.

## Trade-offs (re-evaluated)

- **Query cache bypass**: master's `resolve_data_source_name` runs through `select_one`, so a repeated describe within a single request can be served from the AR query cache. NAME_RESOLVE always hits the server. In practice, Rails' schema cache sits above `resolve_data_source_name` and caches `columns` / `indexes` / `data_source_exists?` at a higher level, so warm production paths don't repeatedly invoke this method. Boot-time schema-cache population (where it is called repeatedly) runs without the query cache middleware active anyway.
- **Consistency**: every other lookup in this adapter reads from the `all_*` dictionaries. With #2513 already calling `DBMS_METADATA.GET_DDL`, this is no longer a style break.
- **Readability**: an anonymous PL/SQL block is harder to read than a SQL query. Outweighed by the speedup, given that #2513's PL/SQL is already in the codebase.

## Test plan

- [x] `bundle exec rspec` passes on CRuby 4.0.2 / Oracle 23ai (453 examples, 0 failures, 6 pending)
- [x] `bundle exec rspec` passes on JRuby 10.0.5.0 / Oracle 23ai (458 examples, 0 failures, 6 pending)
- [x] Instrumentation regression spec from #2545 stays green (SCHEMA event fires)
- [x] Circular synonym specs updated to match NAME_RESOLVE's `ORA-00980` surfacing
- [x] Mixed-case quoted-owner regression spec
- [x] Benchmark script (`script/benchmark_describe.rb`) targets `resolve_data_source_name`

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/jdbc_connection.rb
@@ -306,6 +306,25 @@ module ActiveRecord
           @database_version ||= (md = raw_connection.getMetaData) && [md.getDatabaseMajorVersion, md.getDatabaseMinorVersion]
         end
 
+        # Resolve a schema object (table, view, synonym) to [owner, object_name]
+        # in a single round trip via DBMS_UTILITY.NAME_RESOLVE. Private and
+        # public synonyms are chased server-side, and Oracle raises ORA-01775
+        # natively on a looping chain.
+        def name_resolve(name)
+          cs = @raw_connection.prepareCall("BEGIN DBMS_UTILITY.NAME_RESOLVE(?, 0, ?, ?, ?, ?, ?, ?); END;")
+          cs.setString(1, name)
+          cs.registerOutParameter(2, java.sql.Types::VARCHAR) # schema
+          cs.registerOutParameter(3, java.sql.Types::VARCHAR) # part1 (object name)
+          cs.registerOutParameter(4, java.sql.Types::VARCHAR) # part2
+          cs.registerOutParameter(5, java.sql.Types::VARCHAR) # dblink
+          cs.registerOutParameter(6, java.sql.Types::NUMERIC) # part1_type
+          cs.registerOutParameter(7, java.sql.Types::NUMERIC) # object_number
+          cs.execute
+          [cs.getString(2), cs.getString(3)]
+        ensure
+          cs&.close
+        end
+
         class Cursor
           def initialize(connection, raw_statement, exec_sql = nil)
             @raw_connection = connection
--- a/lib/active_record/connection_adapters/oracle_enhanced/oci_connection.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/oci_connection.rb
@@ -109,6 +109,31 @@ module ActiveRecord
           Cursor.new(self, @raw_connection.parse(sql))
         end
 
+        # Resolve a schema object (table, view, synonym) to [owner, object_name]
+        # in a single round trip via DBMS_UTILITY.NAME_RESOLVE. Private and
+        # public synonyms are chased server-side, and Oracle raises ORA-01775
+        # natively on a looping chain.
+        def name_resolve(name)
+          cursor = @raw_connection.parse(<<~PLSQL)
+            DECLARE
+              l_part2   VARCHAR2(128);
+              l_dblink  VARCHAR2(128);
+              l_type    NUMBER;
+              l_obj_num NUMBER;
+            BEGIN
+              DBMS_UTILITY.NAME_RESOLVE(:name, 0, :out_schema, :out_name,
+                                        l_part2, l_dblink, l_type, l_obj_num);
+            END;
+          PLSQL
+          cursor.bind_param(":name", name)
+          cursor.bind_param(":out_schema", nil, String, 128)
+          cursor.bind_param(":out_name", nil, String, 128)
+          cursor.exec
+          [cursor[":out_schema"], cursor[":out_name"]]
+        ensure
+          cursor&.close
+        end
+
         class Cursor
           def initialize(connection, raw_cursor)
             @raw_connection = connection
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -735,57 +735,49 @@ module ActiveRecord
           end
 
           # Resolves an Oracle data-source name to its underlying [owner, table_name]
-          # by following synonyms through the catalog. Defaults the schema to
-          # `_connection.owner` (the adapter's configured default schema, taken
-          # from `config[:schema]` or `config[:username]`) when the name is not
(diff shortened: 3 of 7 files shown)


yahonda:
> Query cache bypass: master's resolve_data_source_name runs its
catalog query through select_one, which participates in the
ActiveRecord query cache -- a repeated describe in the same request
can be effectively free. NAME_RESOLVE always hits the server.
The sql.active_record event is still emitted, so logging /
subscribers see the call, but there is no cache layer in front of
it.

It should be a concern.

yahonda:
Merging #2521 as the primary direction for the #2429 describe() latency fix.

Reconsideration criterion for this PR: **a concrete report of a synonym-heavy workload where the measured speedup clearly outweighs the code-path complexity** of introducing a `DBMS_UTILITY` PL/SQL call (an anonymous PL/SQL block on OCI, `CallableStatement` on JDBC -- a style break from the rest of the adapter which reads from `ALL_*` dictionaries). Absent that report, the added surface isn't worth carrying.

Keeping this draft open purely as a reference: the rebased code compiles and passes full rspec on both CRuby and JRuby, and the 23c / 11g benchmark tables above document the shape of the speedup if the question ever comes up again.

yahonda:
Closing as superseded by #2686, which landed the production implementation (with the OCI/JDBC OUT-bind code, the NAME_RESOLVE strict-identifier validation, and the spec coverage on top of #2687 / #2688 / #2690's now-merged cleanup helpers).

Branch `resolve-data-source-name-via-name-resolve` and the `script/benchmark_describe.rb` stay around as the historical POC + benchmark reference (the merged gem ships only the adapter changes, not the benchmark script).

PR #2569, opened 2026-04-21
Allow ORACLE_SYSTEM_USER to override hard-coded SYSTEM user

### LINKED ISSUE #2257: System user other than "SYSTEM"

yahonda:
## Summary

- Lets `DatabaseTasks#create` read `ENV["ORACLE_SYSTEM_USER"]` and use it in place of the hard-coded `"SYSTEM"` when establishing the bootstrap connection that runs `CREATE USER` / `GRANT`. Motivation: Oracle Autonomous Database (OCI) ships with `ADMIN` — not `SYSTEM` — as the predefined administrative user, so `rake db:create` on ADB is currently unusable. Oracle docs: [ADMIN User on Autonomous AI Database](https://docs.oracle.com/en-us/iaas/autonomous-database-serverless/doc/autonomous-admin-user-roles.html) ("the predefined administrative user is ADMIN and this account has privileges to manage users and to manage the database").
- Uses `.presence` so an exported-but-empty / whitespace-only value falls back to `"SYSTEM"` instead of attempting a blank login. Validates the resolved username against the existing `ORACLE_IDENTIFIER` regex for consistency with the `:username` and `permissions` validation added in #2558 — the value does not currently flow into SQL (only into `establish_connection`), but fail-fast validation prevents a future refactor from silently creating an injection site.
- Prompt text interpolates the effective username, so the default path prints exactly the same string as today (`"Please provide the SYSTEM password ..."`) and the OCI path prints `"Please provide the ADMIN password ..."`.

Revives the intent of #2261 (closed, targeted `release70`); renewed request in https://github.com/rsim/oracle-enhanced/pull/2261#issuecomment-4290991944. Resolves #2257.

## Test plan

- [x] `bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/database_tasks_spec.rb` — 19 examples, 0 failures. New `describe "create with ORACLE_SYSTEM_USER"` block covers: default SYSTEM, ADMIN override, empty-string fallback, whitespace-only fallback, invalid-identifier rejection, default prompt wording preserved, override prompt wording.
- [x] `bundle exec rubocop` — clean.
- [x] End-to-end integration test on local Oracle 23 Free: created an ADMIN-role user as `SYSTEM`, granted DBA, set `ORACLE_SYSTEM_USER=<that user>` + `ORACLE_SYSTEM_PASSWORD`, and ran `DatabaseTasks.create` — verified the target user was created with all expected privileges and could log in.
- [x] Reviewed with GitHub Copilot CLI.

Co-authored with Joey Perry (@ioev), original author of #2261.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/database_tasks.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/database_tasks.rb
@@ -26,13 +26,21 @@ module ActiveRecord
               raise ArgumentError, "Invalid Oracle privilege in OracleEnhancedAdapter.permissions: #{permission.inspect}"
             end
           end
+          # Oracle Autonomous Database (OCI) ships with ADMIN -- not SYSTEM --
+          # as the predefined administrative user, so the caller needs a way
+          # to override the default. See
+          # https://docs.oracle.com/en-us/iaas/autonomous-database-serverless/doc/autonomous-admin-user-roles.html
+          system_username = ENV["ORACLE_SYSTEM_USER"].presence || "SYSTEM"
+          unless system_username.match?(ORACLE_IDENTIFIER)
+            raise ArgumentError, "Invalid Oracle identifier for ORACLE_SYSTEM_USER: #{system_username.inspect}"
+          end
           quoted_password = %("#{configuration_hash[:password].to_s.gsub('"', '""')}")
 
           system_password = ENV.fetch("ORACLE_SYSTEM_PASSWORD") {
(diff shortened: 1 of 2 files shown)


davidhooey:
Awesome. Thanks!

PR #2594, opened 2026-04-24
Add deferrable foreign key support

yahonda:
## Summary

Brings `deferrable: :immediate | :deferred` to Oracle, following [rails/rails#41487](https://github.com/rails/rails/pull/41487) and the stricter validation enforced after [rails/rails@c09956a3add9](https://github.com/rails/rails/commit/c09956a3add9) removed support for `deferrable: true` in Rails 7.2.

Migrations can now declare deferrable FKs against Oracle:

```ruby
add_foreign_key :aliases, :people, deferrable: :deferred
```

and `foreign_keys(table_name)` / schema dumps round-trip the option.

Oracle accepts [`SET CONSTRAINTS [ALL | name] { IMMEDIATE | DEFERRED }`](https://docs.oracle.com/en/database/oracle/oracle-database/26/sqlrf/SET-CONSTRAINTS.html) within a transaction to flip the check mode of deferrable constraints at runtime.

## Changes

- `supports_deferrable_constraints?` returns `true`.
- `OracleEnhanced::SchemaCreation#visit_ForeignKeyDefinition` appends ``DEFERRABLE INITIALLY { IMMEDIATE | DEFERRED }``; this covers both `ALTER TABLE … ADD CONSTRAINT` and inline FKs in `create_table`.
- `foreign_keys(table_name)` also selects `ALL_CONSTRAINTS.DEFERRABLE` / `DEFERRED` and exposes `options[:deferrable]` as `:immediate` / `:deferred` / `nil`.
- `add_foreign_key` validates the option via `assert_valid_deferrable`, raising `ArgumentError` on anything other than `nil` / `false` / `:immediate` / `:deferred` — matching the PostgreSQL adapter after [rails/rails@c09956a3add9](https://github.com/rails/rails/commit/c09956a3add9).
- `structure_dump.foreign_key_definition` emits the same clause in structure dumps.
- Specs cover both DDL + introspection and schema-dump round-trip.

## Test plan

- [x] `bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/schema_statements_spec.rb -e "deferrable"` (4 examples)
- [x] `bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/schema_dumper_spec.rb -e "deferrable"` (3 examples)
- [x] `bundle exec rspec … -e "foreign"` — full FK suites green (30 examples total, no regressions)
- [ ] CI against supported Oracle versions

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -541,6 +541,12 @@ module ActiveRecord
           SQL
         end
 
+        def add_foreign_key(from_table, to_table, **options)
+          assert_valid_deferrable(options[:deferrable])
+
+          super
+        end
+
         # get table foreign keys for schema dump
         def foreign_keys(table_name) # :nodoc:
           (_owner, desc_table_name) = resolve_data_source_name(table_name)
@@ -551,6 +557,8 @@ module ActiveRecord
                   ,cc.column_name
                   ,c.constraint_name name
                   ,c.delete_rule
+                  ,c.deferrable
+                  ,c.deferred
               FROM all_constraints c, all_cons_columns cc,
                    all_constraints r, all_cons_columns rc
              WHERE c.owner = SYS_CONTEXT('userenv', 'current_schema')
@@ -573,6 +581,7 @@ module ActiveRecord
               primary_key: oracle_downcase(row["references_column"])
             }
             options[:on_delete] = extract_foreign_key_action(row["delete_rule"])
+            options[:deferrable] = extract_foreign_key_deferrable(row["deferrable"], row["deferred"])
             ActiveRecord::ConnectionAdapters::ForeignKeyDefinition.new(oracle_downcase(table_name), oracle_downcase(row["to_table"]), options)
           end
         end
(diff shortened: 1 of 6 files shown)


lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
```diff
@@ -646,6 +655,17 @@ def extract_foreign_key_action(specifier)
             end
           end
 
+          def extract_foreign_key_deferrable(deferrable, deferred)
+            return unless deferrable == "DEFERRABLE"
```
yahonda: `ALL_CONSTRAINTS.DEFERRABLE` is `NOT DEFERRABLE` for ordinary FKs, so returning `nil` here loses the non-deferrable state. After advertising `supports_deferrable_constraints?`, Rails expects `foreign_keys(...).first.deferrable` to round-trip as `false` for that case (that is what PostgreSQL returns, and what the upstream foreign-key tests assert). On this branch, `add_foreign_key(..., deferrable: false)` creates the right DDL but introspects back as `nil`, so the adapter now diverges from Rails/PG semantics and the new spec at lines 755-760 codifies that mismatch. Mapping `NOT DEFERRABLE` to `false` here would keep Oracle aligned with the existing Active Record contract.`

yahonda (commented):
One correctness issue in the Oracle FK introspection path.

lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
yahonda: Good catch — fixed in 955d817e. `extract_foreign_key_deferrable` now returns `false` for `NOT DEFERRABLE` (mirroring PG’s `deferrable && (deferred ? :deferred : :immediate)` which short-circuits to `false` when the boolean is false), and the corresponding spec asserts `be(false)` so round-tripping matches the AR contract.

PR #2596, opened 2026-04-25
Phase 2: Auto-enable identity primary keys in Migration[8.2]+ (depends on Rails per-adapter-migration-compatibility) [DRAFT]

yahonda:
## Summary

**Phase 2** of splitting #2579: build on Phase 1's explicit `identity:`
option (now merged) and flip the **default** for `create_table` to
`identity: true` on Oracle 12.1+, scoped to migrations declared at
`Migration[8.2]` or later. `Migration[8.1]` and earlier, `Schema.define`,
and direct `connection.create_table` keep the existing sequence + trigger
behavior, so applications see no change unless they explicitly adopt
`Migration[8.2]`.

**This PR is a draft because it depends on the unmerged Rails extension
point** `AbstractAdapter#migration_compatibility_module_for(migration_class)`
plus the `Migration::Compatibility::Versioned` convention from
[yahonda/rails `per-adapter-migration-compatibility`](https://github.com/yahonda/rails/tree/per-adapter-migration-compatibility).
Once that lands on `rails/rails#main`:
1. Switch `Gemfile` back to `rails/rails#main`
2. Mark this PR ready for review

### Behavior matrix

| Caller | `identity:` arg | Oracle server | Result |
|---|---|---|---|
| `Migration[8.2]+` | omitted | ≥ 12.1 | Identity (V8_2 default) |
| `Migration[8.2]+` | omitted | < 12.1 | Sequence (V8_2 auto-inject gated by `supports_identity_columns?`) |
| `Migration[8.2]+` | omitted, with `sequence_name:` / `sequence_start_value:` | any | Sequence (V8_2 treats sequence options as opt-out) |
| `Migration[8.2]+` | `true` | ≥ 12.1 | Identity (explicit) |
| `Migration[8.2]+` | `true` | < 12.1 | `ArgumentError` (Phase 1) |
| `Migration[8.2]+` | `false` | any | Sequence (per-table opt-out) |
| `Migration[8.1]` or older | omitted | any | Sequence (forced by V8_1 module) |
| `Migration[8.1]` or older | `true` | ≥ 12.1 | Identity (explicit still honored) |
| `Migration[8.1]` or older | `true` | < 12.1 | `ArgumentError` |
| `Migration[8.1]` or older | `false` | any | Sequence |
| `Schema.define` / direct `connection.create_table` | omitted | any | Sequence (no migration version context, default unchanged) |
| `Schema.define` / direct `connection.create_table` | `true` | ≥ 12.1 | Identity (explicit) |
| `Schema.define` / direct `connection.create_table` | `false` | any | Sequence |

### Implementation notes

- `OracleEnhanced::MigrationCompatibility` registers two compat modules
  via Rails' `Migration::Compatibility::Versioned`:
  - `V8_2` forces `options[:identity] = true` when the option is omitted,
    no `sequence_name:` / `sequence_start_value:` is given,
    `id: :primary_key` is in effect (default), the PK is not composite,
    and the server supports identity columns. This is the new flip,
    scoped to migrations at `Migration[8.2]+`.
  - `V8_1` forces `options[:identity] = false` for migrations at
    `Migration[8.1]` or earlier. Because Rails' versioned dispatch
    includes both modules for older migrations and V8_1 sets the
    explicit option first, V8_2 sees `options.key?(:identity)` and
    leaves it alone — older migrations keep sequence-backed behavior.
- `OracleEnhancedSchemaStatements#create_table` honors
  `options[:identity]` exactly as given and does **no** auto-injection
  of its own. Callers that bypass migration compat (`Schema.define`,
  direct `connection.create_table`) therefore keep the existing
  sequence-backed default unless they pass `identity: true` explicitly.
- `OracleEnhancedSchemaStatements#rename_table` and `#drop_table` are
  now identity-aware: they no longer rename or drop `<table>_seq` for
  tables whose primary key is an Oracle identity column. Previously,
  an unrelated application-owned sequence sharing that name would be
  silently mutated or removed. `full_drop` already had this protection
  via its `ISEQ$$_` filter; this brings the table-level DDL paths in
  line.
- `OracleEnhancedSchemaDumper` emits `identity: true` / `identity: false`
  explicitly for every primary-key table when the connected server
  supports identity columns, so `db/schema.rb` roundtrips remain
  unambiguous regardless of which loader runs them.
- `OracleEnhancedDatabaseStatements#empty_insert_statement_value` is
  already in master from #2609 — this PR depends on that change so
  identity tables can satisfy `Model.create!` with no attributes.

### Tests

- New `migration_compatibility_spec.rb` cases exercise the V8_2 / V8_1
  layering, the sequence-option opt-out, and the `Schema.define` /
  direct-call path that preserves sequence-backed PKs by default.
- New `identity_primary_key_spec.rb` cases cover identity-aware
  `rename_table` and `drop_table` against an unrelated app-owned
  `<table>_seq`, plus the identity counterparts to the existing
  sequence-prerequisite tests (`pk_and_sequence_for`, `NEXTVAL`-free
  insert, `full_drop` filter, long-name rename).

## Test plan

- [x] Full spec suite on Oracle 23ai (FREEPDB1) — 573 examples, 0 failures, 7 pending
- [ ] CI on full matrix (3.3 / 3.4 / 4.0 / jruby)
- [ ] Switch Gemfile back to `rails/rails#main` once `per-adapter-migration-compatibility` lands

## Related

- Phase 1: #2579 (explicit `identity:` option) — merged
- Phase 2 prerequisite: #2609 (`empty_insert_statement_value`) — merged
- Rails dependency: yahonda/rails `per-adapter-migration-compatibility`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ group :development do
   gem "rspec"
   gem "rdoc"
   gem "rake"
-  gem "activerecord",   github: "rails/rails", branch: "main"
+  gem "activerecord",   github: "yahonda/rails", branch: "per-adapter-migration-compatibility"
   gem "ruby-plsql", github: "rsim/ruby-plsql", branch: "master"
 
   platforms :ruby do
--- a/lib/active_record/connection_adapters/oracle_enhanced/migration_compatibility.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/migration_compatibility.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module ActiveRecord
+  module ConnectionAdapters
+    module OracleEnhanced
+      module MigrationCompatibility # :nodoc: all
+        extend ActiveRecord::Migration::Compatibility::Versioned
+
+        module V8_2
+          def create_table(table_name, **options)
+            options[:identity] = true if oracle_enhanced_default_to_identity?(options)
+            super
+          end
+
+          private
+            def oracle_enhanced_default_to_identity?(options)
+              return false if options.key?(:identity)
+              return false if options[:sequence_name] || options[:sequence_start_value]
+              return false if options.fetch(:id, :primary_key) != :primary_key
+              return false if options[:primary_key].is_a?(Array)
+              connection.supports_identity_columns?
+            end
+        end
+
+        module V8_1
+          def create_table(table_name, **options)
+            options[:identity] = false unless options.key?(:identity)
+            super
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_dumper.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_dumper.rb
@@ -187,6 +187,8 @@ module ActiveRecord # :nodoc:
                   tbl.print ", primary_key_trigger: true"
                   default_name = @connection.default_trigger_name(table).upcase
                   tbl.print ", trigger_name: #{trigger_name.downcase.inspect}" unless trigger_name == default_name
+                elsif @connection.supports_identity_columns?
+                  tbl.print ", identity: false"
                 end
               when Array
                 tbl.print ", primary_key: #{pk.inspect}"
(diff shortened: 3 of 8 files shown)


yahonda:
Rebased onto current master (e8e1677c) to pick up the enforced foreign key support (#2793), Codespaces docs (#2807, #2808), and other recent changes.

yahonda:
Superseded by #2816, which implements identity primary keys for Migration[8.2]+ on the v4 per-adapter CompatibilityBehavior design. Closing in favor of #2816.

PR #2630, opened 2026-04-28
[measurement only] #2626 + #2629 combined

yahonda:
**Not for merge.** Combined branch of #2626 and #2629 to measure whether binding the dictionary-query literals (#2629) cancels out the CI runtime regression introduced by removing the `cursor_sharing = force` default (#2626 alone showed +20-43% on Run RSpec).

If Run RSpec on this branch comes back near master baseline, #2629 then #2626 can land as planned.


yahonda:

PR #2631, opened 2026-04-28
[measurement only] master + prepared_statements: false

yahonda:
Throwaway PR to trigger CI for measurement of (prepared_statements: false × cursor_sharing: force) baseline.


yahonda:

PR #2632, opened 2026-04-28
[measurement only] #2626+#2629 + prepared_statements: false

yahonda:
Throwaway PR to trigger CI for measurement of (prepared_statements: false × cursor_sharing: default + binds) cell.


yahonda:

PR #2633, opened 2026-04-28
[measurement only] #2629 + prepared_statements: false

yahonda:
Throwaway PR to measure #2629 alone with prepared_statements: false. Verifies #2629 doesn't add new failures vs master + ps:false baseline.


yahonda:

PR #2640, opened 2026-04-28
[measurement only] regression spec for #2637 without the lib fix

yahonda:
Throwaway PR to verify the readonly LOB regression specs added in #2637 actually fail on master without the lib fix. Expected outcome: the two new `should respect attr_readonly setting for {C,N}LOB column when prepared_statements is false` examples fail (CLOB readonly + NCLOB readonly), and only those.


yahonda:
Verified: regression specs fail pinpoint (2 failures, no others) without #2637's lib fix. Closing measurement-only PR.

PR #2642, opened 2026-04-28
Skip RETURNING when the caller supplies the primary key value

### LINKED ISSUE #2639: String primary key INSERT raises ORA-01722 on prepared_statements: false
### Background

When a model has a String primary key and is configured with `prepared_statements: false`, `create!` raises `ORA-01722` ("unable to convert string value to a number") on the primary-key column.

Reproducible on master with `ORACLE_ENHANCED_PREPARED_STATEMENTS_FALSE=1` (#2635). The pre-existing spec asserts the correct behavior and fails in that configuration:

```
OracleEnhancedAdapter schema definition primary_key inside create_table block
  with type and keyword options
  inserts via the Rails insert path on a String primary key

  Failure/Error: expect { klass.create!(code: "ABC", name: "alpha") }.not_to raise_error
  expected no Exception, got #<OCIError:
    ORA-01722: unable to convert string value containing 'A' to a number: CODE
    ORA-03302: ... invalid string value: ABC
  > with backtrace:
    # stmt.c:265:in oci8lib_400.so
    # ...oci8/cursor.rb:137:in 'OCI8::Cursor#exec'
    # .../oci_connection.rb:154:in '...Cursor#exec_update'
    # .../database_statements.rb:84:in 'block (2 levels) in _exec_insert'
    # ...
```

`spec/active_record/connection_adapters/oracle_enhanced/schema_statements_spec.rb:142`

### Likely cause

Under `prepared_statements: false`, the value AR passes for the primary-key column ends up being typed against an Integer / Number context somewhere on the `_exec_insert` path. The String value `"ABC"` then fails the implicit conversion to NUMBER. The `:returning_id` bind plumbing in `database_statements.rb` is one suspect — it is wired specifically for an Integer PK and may be applying an Integer type cast to the PK column even when the PK type is String.

### Reproduction

```ruby
ActiveRecord::Schema.define do
  drop_table :things, if_exists: true
  create_table :things, id: false do |t|
    t.primary_key :code, :string, limit: 32
    t.string :name
  end
end

class Thing < ActiveRecord::Base
  self.table_name = "things"
  self.primary_key = "code"
end

Thing.create!(code: "ABC", name: "alpha")
# => ORA-01722 (only on prepared_statements: false connections)
```

`prepared_statements: true` connections do not hit this because the bind path types the primary-key value against the actual column type via OCI binding.

### Related

- #2622 — broader test-coverage gap that allowed this to stay latent
- #2635 — workflow that surfaces this on `prepared_statements: false`
- #2636 / #2637 — readonly LOB on create (separate prepared_statements: false bug)
- (separate issue) — CLOB equality on select

yahonda:
## Summary

Fixes `ORA-01722: unable to convert string value containing 'A' to a number` raised on `prepared_statements: false` connections when a model with a String (VARCHAR2) primary key inserts a record.

## Root cause

`sql_for_insert` appends ` RETURNING <pk> INTO :returning_id` (Integer-typed out-bind) only when the caller did **not** supply the primary key, so Rails can read back a database-generated id from a sequence / trigger. The existing predicate is

```ruby
binds.none? { |bind| bind.name == pk.to_s }
```

— "is the PK absent from the bind list?". On `prepared_statements: true` this works because the caller-supplied value lives in `binds`. On `prepared_statements: false` the caller-supplied value is **interpolated into the column list** of the SQL and `binds` is empty, so the predicate wrongly concludes the PK is auto-generated, the RETURNING is appended, and Oracle raises `ORA-01722` against the Integer out-bind when the PK column is non-numeric.

## Fix

Detect the `prepared_statements: false` case by also looking at the INSERT column list in the SQL. If the PK column is present in either the binds **or** the column list, the user supplied the value and no RETURNING is appended.

```ruby
def insert_supplies_pk?(sql, binds, pk_name)
  return true if binds.any? { |bind| bind.name == pk_name }
  match = sql.match(/\AINSERT\s+INTO\s+\S+\s*\(([^)]*)\)/i)
  return false unless match
  match[1].scan(/[A-Za-z_][\w$#]*/).any? { |col| col.casecmp?(pk_name) }
end
```

`prepared_statements: true` is unaffected — the bind-list branch still short-circuits exactly as before.

## Tests

Adds a regression spec next to the pre-existing String-PK Rails-insert test, with `prepared_statements` forced to `false` via `instance_variable_set` (matching the pattern used by #2485 / #2641 / #2637 readonly / equality specs). Verified failing on master without the fix and passing with it.

The pre-existing top-level `inserts via the Rails insert path on a String primary key` (`schema_statements_spec.rb:142`) — failing on the `test_prepared_statements_false` workflow #2635 — also returns to green.

Closes #2639.


yahonda:
Superseded by #2644 which uses a simpler regex parsing approach. Closing in favor of that.

PR #2643, opened 2026-04-28
[measurement only] regression spec for #2642 without the lib fix

yahonda:
Verify the unprepared_statement-block spec from #2642 actually fails on master without the lib fix. Expected: only the new `inserts via the Rails insert path on a String primary key when prepared_statements is false` example fails (1 failure).


yahonda:
Verified: the regression spec fails on master without the lib fix as expected (1 failure pinpoint). Closing validation PR.

PR #2664, opened 2026-04-30
Restore Arel In and NoIn support

akostadinov:
in #2075 and 94057806bc8a3b8b4efbc4af1e360178e1c25465 the Arel methods `visit_Arel_Nodes_In` and `visit_Arel_Nodes_NotIn` were removed because Rails stopped using them by default but these are still accessible code paths when Arel is directly used.

That can be the user constructing queries with Arel directly or some framework like baby_squeel doing it.

This commit restores the two methods (with a fix-up for sub-queries) and adds tests to make sure low-level Arel usage is still supported in the long-run.

### DIFF
--- a/lib/arel/visitors/oracle_common.rb
+++ b/lib/arel/visitors/oracle_common.rb
@@ -97,6 +97,35 @@ module Arel # :nodoc: all
           end
           array
         end
+
+        def visit_Arel_Nodes_In(o, collector)
+          attr, values = o.left, o.right
+          return super unless values.is_a?(Array)
+
+          in_clause_length = @connection.in_clause_length
+          return super if values.length <= in_clause_length
+
+          # Split into multiple IN nodes and combine with OR
+          in_nodes = values.each_slice(in_clause_length).map do |slice|
+            Arel::Nodes::In.new(attr, slice)
+          end
+          or_node = Arel::Nodes::Or.new(in_nodes)
+          visit(Arel::Nodes::Grouping.new(or_node), collector)
+        end
+
+        def visit_Arel_Nodes_NotIn(o, collector)
+          attr, values = o.left, o.right
+          return super unless values.is_a?(Array)
+
+          in_clause_length = @connection.in_clause_length
+          return super if values.length <= in_clause_length
+
+          # Split into multiple NOT IN nodes and combine with AND
+          not_in_nodes = values.each_slice(in_clause_length).map do |slice|
+            Arel::Nodes::NotIn.new(attr, slice)
+          end
+          visit(Arel::Nodes::And.new(not_in_nodes), collector)
+        end
     end
   end
 end
--- a/spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
+++ b/spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb
@@ -244,7 +244,7 @@ describe "OracleEnhancedAdapter" do
       class ::TestPost < ActiveRecord::Base
         has_many :test_comments
       end
-      @ids = (1..1010).to_a
+      @ids = (1..2010).to_a
       TestPost.transaction do
         @ids.each do |id|
           TestPost.create!(id: id, title: "Title #{id}")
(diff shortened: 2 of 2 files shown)


yahonda (commented):
Thanks for picking this up — the underlying problem is real (direct Arel use via `baby_squeel` etc. still produces `Arel::Nodes::In` / `Arel::Nodes::NotIn`, which after #2075 fall through to upstream `Arel::Visitors::ToSql` and hit ORA-01795 on >1000 items).

### `collector.preparable = false` is missing on the Array branch

Upstream `Arel::Visitors::ToSql#visit_Arel_Nodes_In` / `#visit_Arel_Nodes_NotIn` set `collector.preparable = false` whenever `values` is an `Array`. The pattern was tightened in [rails/rails#51497](https://github.com/rails/rails/pull/51497) to the exact form this pull request should mirror — `preparable = false` for Arrays, but **not** for subselects (so subselect IN can stay in the statement cache):

```ruby
# rails/rails#51497
def visit_Arel_Nodes_In(o, collector)
  attr, values = o.left, o.right

  if Array === values
    collector.preparable = false
    ...
```

This pull request #2664 currently leaves `preparable = true` on every path — verified locally against master branch, where Arrays correctly produce `preparable = false` today and pull request #2664 changes that to `true`.

The minimal change is one line per method, added right after the `super` short-circuit:

```ruby
def visit_Arel_Nodes_In(o, collector)
  attr, values = o.left, o.right
  return super unless values.is_a?(Array)

  collector.preparable = false   # <-- add this
  ...
```

```ruby
def visit_Arel_Nodes_NotIn(o, collector)
  attr, values = o.left, o.right
  return super unless values.is_a?(Array)

  collector.preparable = false   # <-- add this
  ...
```

---

Could you address this one before we move forward? The other notes I had — moving the visitor-level assertions under `spec/.../arel/oracle{,12}_spec.rb` to match the layout settled in #2660 / #2661 / #2662, and adding edge-case tests (empty array, exactly 1000, exactly 1001, `unboundable`-only) — I'd rather handle in a follow-up pull request myself than rewrite a lot of what you've put together here, and that test-layout work is in flight on my side already.

yahonda:
Also, please squash the commits before pushing the fix — this repo merges PRs as a single commit, so a clean linear history on the branch keeps the merge tidy.

akostadinov:
@yahonda thinking that through, I thought it may make more sense to deconstruct the original `Arel::Nodes::In` / `Arel::Nodes::NotIn` into multiple `In`/`NotIn`/`And`/`Or` nodes and delegate to the Rails default logic for better future proofing. Repeating the implementation exposes us to missing future changes.

This can perhaps also be applied to `visit_Arel_Nodes_HomogeneousIn`.

But in case you prefer the original approach and just add the `collector.preparable = false` line, let me know and I will submit that instead.

yahonda:
Hi,

> I thought it may make more sense to deconstruct the original Arel::Nodes::In / Arel::Nodes::NotIn into multiple In/NotIn/And/Or nodes and delegate to the Rails default logic for better future proofing. Repeating the implementation exposes us to missing future changes.

This approach should be fine.

akostadinov:
Ok, great! Just to be sure I'm not missing something, let me know if you expect further changes to this PR from me before merging. My understanding is that this PR is complete at this point.

akostadinov:
FYI I notice some strange issue, I will debug it the coming week. So maybe not rush to merge.

akostadinov:
Ok, good to go, there was no actual bug. But in older Rails the `Or` node received two parameters while now it can receive an array. So it failed for old Rails which is no concern for this PR.

But I simplified a little more how the `And` and `Or` nodes are generated, now with just `Arel::Nodes::Or.new(in_nodes)` instead the nesting with `reduce` previously done. Same for the `And` node. I think this is much better.

Another small change from last version is making the list 2010 elements long to make sure that 3 chunks will work as good as 2 chunks to catch any regressions in handling more than 2 chunks of elements.

yahonda:
Thanks for the update.

yahonda:
Oracle enhanced adapter 8.1.4 has been released that includes this fix.

PR #2671, opened 2026-05-01
[WIP] Oracle12 default-visitor regression: FIRST_VALUE+DISTINCT+ORDER BY drops alias_N__ rewrite

yahonda:
## Status

Draft. **Red** stage of a TDD-style red/green pair:

- This PR (red): adds a failing-by-design unit test that documents the regression exposed by #2573.
- Follow-up PR (green): fixes the regression so the test passes.

## Why

#2573 makes `Arel::Visitors::Oracle12` the default visitor for every connection (previously the default was effectively `Arel::Visitors::Oracle` because `supports_fetch_first_n_rows_and_offset?` was hardcoded to `false`).

`Arel::Visitors::Oracle#order_hacks` rewrites the outer `ORDER BY` of a DISTINCT query whose projection list contains `FIRST_VALUE(...) AS alias_0__` so the ORDER BY references the `alias_0__` column instead of the inner expression. `Arel::Visitors::Oracle12` does **not** define `order_hacks`, so once it becomes the default visitor the rewrite drops out and Oracle raises **ORA-01791 ("not a SELECTed expression")** at execute time.

`columns_for_distinct` is what produces the `FIRST_VALUE(...) AS alias_0__` projection, so any AR query that goes through that helper (e.g. `Model.distinct.order(:created_at)` in shapes that AR materialises with a window function) is potentially affected.

## What this PR adds

A new describe block in `spec/active_record/connection_adapters/oracle_enhanced/columns_for_distinct_spec.rb`:

```
describe "integration with default `@conn.visitor` (Oracle12)" do
```

— mirrors the three assertions in the existing `"integration with Arel::Visitors::Oracle#order_hacks"` block but routes through `@conn.visitor` (= the default visitor on master after #2573) instead of a manually-constructed `Arel::Visitors::Oracle`. All three currently fail:

```
1) ... preserves DESC NULLS LAST in the outer ORDER BY for a raw SQL order
   expected "... ORDER BY posts.created_at DESC NULLS LAST"
   to include "ORDER BY alias_0__ DESC NULLS LAST"

2) ... preserves DESC NULLS LAST in the outer ORDER BY for an Arel ordering node
   expected "... ORDER BY posts.created_at DESC NULLS LAST"
   to include "ORDER BY alias_0__ DESC NULLS LAST"

3) ... preserves NULLS FIRST without emitting DESC for an ASC-with-nulls order
   expected "... ORDER BY posts.created_at ASC NULLS FIRST"
   to match /ORDER BY alias_0__ NULLS FIRST\b/
```

The existing `"integration with Arel::Visitors::Oracle#order_hacks"` block stays as it was — pinned to a manually-constructed Oracle visitor — so the file is intentionally a mix of green and red.

## What this PR does NOT do

It does not fix the regression. The likely fix is to hoist `order_hacks` and `split_order_string` from `Arel::Visitors::Oracle` into `Arel::Visitors::OracleCommon` (the module both visitors include) and call `o = order_hacks(o)` from `Arel::Visitors::Oracle12#visit_Arel_Nodes_SelectStatement` before delegating to `super`. That fix lives in a follow-up PR; this one is just the failing test that locks the contract.

## Why a separate PR

- Keeps #2573's scope tight on the `limit + lock` `ArgumentError` removal.
- Makes the regression visible to CI and reviewers without entangling it with a fix that may need its own discussion (e.g. how generic `order_hacks` should be in `OracleCommon`, whether the rewrite should be opt-in, what to do if the user is on `use_old_oracle_visitor = true`).
- Lets us measure how often the Oracle12-without-order_hacks path is actually hit in real applications before deciding the urgency of the green PR — the red examples here are visitor-compile assertions (no live DB). If the wider integration suite turns this into ORA-01791 in many places, the green PR is urgent; if not, this stays open as a known-issue marker for a while.

## Refs

- #2573 — the PR that exposes the regression (this branch is forked off `fix-2395-fetch-first-n-rows`)
- The regression was flagged by Copilot CLI during review of #2573.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -190,10 +190,11 @@ module ActiveRecord
 
       ##
       # :singleton-method:
-      # By default, OracleEnhanced adapter will use Oracle12 visitor
-      # if database version is Oracle 12.1.
-      # If you wish to use Oracle visitor which is intended to work with Oracle 11.2 or lower
-      # for Oracle 12.1 database you can add the following line to your initializer file:
+      # By default, OracleEnhanced adapter uses `Arel::Visitors::Oracle12`,
+      # which emits `FETCH FIRST n ROWS ONLY` / `OFFSET n ROWS` for limits
+      # and offsets. To opt back into `Arel::Visitors::Oracle` (ROWNUM-based
+      # output, intended for pre-12c servers) on every connection, add the
+      # following line to your initializer:
       #
       #   ActiveRecord::ConnectionAdapters::OracleEnhancedAdapter.use_old_oracle_visitor = true
       cattr_accessor :use_old_oracle_visitor
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -353,12 +353,7 @@ module ActiveRecord
       end
 
       def supports_fetch_first_n_rows_and_offset?
-        false
-
-        # TODO: At this point the connection is not initialized yet,
-        # so `database_version` raises an error
-        #
-        # !use_old_oracle_visitor && database_version.first >= 12
+        !use_old_oracle_visitor
       end
 
       def supports_datetime_with_precision?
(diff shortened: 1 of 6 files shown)


yahonda:
Subsumed by #2573, which now hoists `order_hacks` and `split_order_string` into `Arel::Visitors::OracleCommon` so `Arel::Visitors::Oracle12` (the new default after #2573) also picks up the rewrite. The failing examples added here are covered there by the renamed `integration with order_hacks` describe block in `columns_for_distinct_spec.rb`, which now routes through `@conn.visitor` (= the default visitor) and exercises the same shapes.

PR #2680, opened 2026-05-01
Cap :schema length at the connected DB's identifier limit (closes #2561)

### LINKED ISSUE #2561: configure_connection: add version-aware length cap on SCHEMA_IDENTIFIER_PATTERN
## Context

PR #2557 added `SCHEMA_IDENTIFIER_PATTERN = /\A[[:alpha:]][\w$#]*\z/` to validate the `:schema` connection option before interpolating it into `ALTER SESSION SET CURRENT_SCHEMA`. The regex matches the Oracle unquoted-identifier grammar for characters but deliberately omits a length bound, because:

- Oracle 12.1 and earlier: identifiers capped at 30 bytes.
- Oracle 12.2 and later: identifiers can be up to 128 bytes.

Using `{0,29}` (the bound the adapter's existing `OracleEnhanced::Quoting::NONQUOTED_OBJECT_NAME` applies) would wrongly reject legal 12.2+ identifiers; using `{0,127}` would over-accept on 12.1. The PR opted to skip the cap entirely and let Oracle reject an oversized identifier with `ORA-00972` at `ALTER SESSION` time.

## Problem

This means a 1000-character `:schema` value passes the adapter's pre-check and only fails once Oracle sees the ALTER SESSION. The upfront `ArgumentError` that the new validation promises is slightly weaker as a result.

## Suggested follow-up

Pick a bound that depends on the connected Oracle's maximum identifier length (e.g., query once post-connect, cache on the adapter) and reject lengths beyond that in `configure_connection` so the user gets an `ArgumentError` with a clear message rather than an opaque Oracle error.

Raised from the Copilot review on #2557; left as-is for that PR per the stated design choice.

yahonda:
Closes #2561.

#2557 added `SCHEMA_IDENTIFIER_PATTERN = /\A[[:alpha:]][\w$#]*\z/` to validate the `:schema` connection option's character set before interpolating it into `ALTER SESSION SET CURRENT_SCHEMA`. The PR deliberately omitted a length cap because the legal cap is 30 bytes on Oracle 12.1 and earlier and 128 bytes on 12.2+; a single hard-coded bound would either reject legal 12.2+ identifiers or over-accept on 12.1.

Now that `database_version` is a Comparable `Version` object (#2677) and `supports_longer_identifier?` reflects the connected DB's pure capability (#2576: true on 12.2+, false otherwise), the cap can be added cleanly: reject in `configure_connection` when `schema.bytesize > (supports_longer_identifier? ? 128 : 30)`, with an `ArgumentError` naming both the limit and the actual byte count.

## Why `supports_longer_identifier?` and not `max_identifier_length`

The cap should track *Oracle's actual schema-name capability*, not the adapter's emit policy:

- `supports_longer_identifier?` is the pure DB capability flag — true on 12.2+, false otherwise.
- `max_identifier_length` is config-resolved and tracks adapter-emitted identifiers (table names, column names, etc.). With `identifier_max_length: :short` or the deprecated `use_shorter_identifier = true`, it returns 30 even on a 12.2+ server.

`:schema` references a *pre-existing* Oracle USER, not an identifier the adapter emits. A user might legitimately have `identifier_max_length: :short` (because they want adapter-emitted identifiers to stay 30-byte for downstream compatibility) and still be connecting to a 50-byte schema/user that exists in the DB. Anchoring the validation to `max_identifier_length` would falsely reject that case; anchoring to `supports_longer_identifier?` matches what Oracle would actually accept on the wire.

(An earlier revision of this PR used `max_identifier_length` for "adapter consistency"; reviewers correctly flagged that the consistency argument doesn't apply when the identifier in question is something the adapter is *referencing*, not *emitting*.)

## Tests

- `should raise ArgumentError for a :schema value longer than the connected DB's identifier limit` — exercises the cap with `cap + 1` ASCII bytes. The assertion reads the cap dynamically (`conn.supports_longer_identifier? ? 128 : 30`), so the example passes on every CI matrix row.
- `rejects a multibyte :schema name by bytesize, not character count` — uses `é` (2 UTF-8 bytes) as the first character followed by `cap` ASCII characters, producing `cap + 1` chars but `cap + 2` bytes. The `SCHEMA_IDENTIFIER_PATTERN` regex's `[[:alpha:]]` admits non-ASCII letters as the first char, so this exercises the bytesize-vs-length distinction on the byte side.

## Test plan

- [x] `bundle exec rspec` on Oracle 23ai — 705 examples, 0 failures, 11 pending.
- [x] `bundle exec rubocop` — clean.
- [ ] CI green on all matrix rows.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb
@@ -1053,6 +1053,17 @@ module ActiveRecord
           unless schema.match?(SCHEMA_IDENTIFIER_PATTERN)
             raise ArgumentError, "Invalid :schema value #{schema.inspect}; must be an Oracle unquoted identifier"
           end
+          # The cap is the *connected database*'s capability, not the adapter's
+          # `max_identifier_length` (which is config-resolved and tracks
+          # adapter-emitted identifiers; `identifier_max_length: :short` and
+          # `use_shorter_identifier = true` would wrongly reject existing
+          # 31..128-byte schema names on a 12.2+ server).
+          schema_byte_limit = supports_longer_identifier? ? 128 : 30
+          if schema.bytesize > schema_byte_limit
+            raise ArgumentError,
+              "Invalid :schema value #{schema.inspect}; exceeds the #{schema_byte_limit}-byte " \
+              "identifier limit (#{schema.bytesize} bytes)."
+          end
           execute("alter session set current_schema = #{schema}", "SCHEMA")
         end
 
--- a/spec/active_record/connection_adapters/oracle_enhanced/connection_spec.rb
+++ b/spec/active_record/connection_adapters/oracle_enhanced/connection_spec.rb
@@ -212,6 +212,29 @@ describe "OracleEnhancedConnection" do
       expect { ActiveRecord::Base.lease_connection }.to raise_error(ArgumentError, /Invalid :schema value/)
     end
 
+    it "should raise ArgumentError for a :schema value longer than the connected DB's identifier limit" do
+      conn = ActiveRecord::Base.connection
+      cap = conn.supports_longer_identifier? ? 128 : 30
+      ActiveRecord::Base.remove_connection
+      overlong = "a" + "b" * cap   # cap + 1 bytes, all valid unquoted-identifier characters
+      ActiveRecord::Base.establish_connection(CONNECTION_PARAMS.merge(schema: overlong))
+      expect { ActiveRecord::Base.lease_connection }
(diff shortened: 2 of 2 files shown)


yahonda:
Closing as wontfix; the upfront bytesize check turns out to be redundant.

`:schema` references an existing Oracle USER, and Oracle itself enforces the identifier-length cap at `CREATE USER` time — so any `:schema` value that corresponds to a real DB user is by definition within the limit. Out-of-range strings (typo / garbage) get rejected by Oracle with `ORA-00972 identifier is too long` at `ALTER SESSION` parse time, which is sufficiently clear without an adapter-side translation.

Issue #2561 closed in the same direction.

PR #2710, opened 2026-05-07
Phase 2: flip add_index unique: true default in Migration[8.2]+ [DRAFT — depends on Rails per-adapter-migration-compatibility]

### LINKED ISSUE #2702: Deprecate implicit UNIQUE CONSTRAINT creation by add_index :col, unique: true
## Background

Since #2610 / #2612, `add_index :t, :col, unique: true, name: :n` emits both a `CREATE UNIQUE INDEX` and a trailing `ALTER TABLE ADD CONSTRAINT n UNIQUE (col) USING INDEX n` for non-functional indexes. This adapter-specific behavior was originally introduced because Oracle requires a UNIQUE/PK CONSTRAINT (not just a UNIQUE INDEX) for FK targetability — without it, a `add_foreign_key` against an indexed-only column hits `ORA-02270`.

Now that #2701 has landed `add_unique_constraint` / `t.unique_constraint` / `remove_unique_constraint`, users have an explicit DSL for the constraint side. The implicit-constraint creation in `add_index` has become redundant and at odds with both Rails core's general convention (PostgreSQL / MySQL / SQLite all create only the index) and the principle of least surprise (`add_index` should add an index, not also a constraint).

## Proposal

Three-phase deprecation:

### Phase 1 — deprecation warning + global opt-out flag (no Rails core dependency) — delivered in #2709

Emit an `ActiveSupport::Deprecation` warning when `add_index` enters the implicit-constraint branch, e.g.:

> `add_index :t, :col, unique: true` creates an implicit named UNIQUE constraint on Oracle. This implicit-constraint behavior will be removed in oracle-enhanced X.0. For an explicit constraint, call `add_unique_constraint :t, :col, name: :n`. For FK targetability without changing index identity, call both:
>   `add_index :t, :col, unique: true`
>   `add_unique_constraint :t, :col`

Add a global opt-out for users who already prefer index-only:

```ruby
# config/initializers/oracle_enhanced.rb
ActiveRecord::ConnectionAdapters::OracleEnhancedAdapter.add_index_unique_creates_constraint = false
```

- `false` → `add_index unique: true` only emits `CREATE UNIQUE INDEX`; no warning
- `true` (Phase 1 default) → current behavior + warning
- `nil` (default) → equivalent to `true`

### Phase 2 — `Migration[N+1]` default flip (depends on Rails `another-33269`)

Following the precedent in #2596, register a `MigrationCompatibility::Versioned` module so that:

- `Migration[N]` and earlier keep current implicit-constraint behavior
- `Migration[N+1]` and later default to index-only (matching the global flag's `false` value)

`Schema.define` and direct `connection.create_table` keep the flag's value as their source of truth, since they have no migration-version context.

### Phase 3 — flag and implicit-constraint code path removal

After Phase 2 has had time to settle (one major release on the new default), remove the transition machinery entirely:

- `OracleEnhancedAdapter.add_index_unique_creates_constraint` cattr_accessor
- `warn_implicit_unique_constraint_deprecation` private helper
- `add_inline_unique_constraints` post-`super` drain in `create_table`
- The trailing `ALTER TABLE ... ADD CONSTRAINT ... USING INDEX` emission in `add_index`
- The `OracleEnhanced.deprecator.silence` wrap added to `schema_define` in spec_helper
- The `MigrationCompatibility` module from Phase 2 (no longer needed once the only behavior is "index only")

End state: `add_index :col, unique: true` only emits `CREATE UNIQUE INDEX` on Oracle, achieving full parity with the Rails-core PostgreSQL / MySQL / SQLite adapters. Callers who need a constraint reach for `add_unique_constraint` (which has been the explicit path since #2701).

This phase has no Rails-core dependency; it can ship one major release after Phase 2.

## Out of scope

- Removing the `add_inline_unique_constraints` post-`super` loop in `create_table` until Phase 2 lands. The deprecation warning fires from the same code path; the implementation stays.
- Auto-rewriting historical migrations.

## Related

- #2701 added `add_unique_constraint` (Closes #2613). Released the explicit-constraint DSL.
- #2596 established the `Migration::Compatibility::Versioned` precedent (depends on yahonda/rails `another-33269`).

## Acceptance criteria

- Phase 1 — deprecation warning + global flag, with specs covering on/off/default and that the warning does not fire when `add_unique_constraint` is called explicitly.
- Phase 2 — `MigrationCompatibility` module registering `V_<next>` that flips the default. Specs covering `Migration[N+1]+`, `Migration[N]` and earlier, and `Schema.define`. (Blocked by `another-33269`.)
- Phase 3 — flag, helper, and implicit-constraint code path are removed. `add_index unique: true` emits only `CREATE UNIQUE INDEX`. Specs cover the new clean behavior; the `schema_define` deprecator silence wrap is no longer needed.

yahonda:
## Summary

Closes #2702 (Phase 2).

**This PR is a draft because it depends on the unmerged Rails extension point** `AbstractAdapter#migration_compatibility_module_for(migration_class)` plus the `Migration::Compatibility::Versioned` convention from [yahonda/rails `per-adapter-migration-compatibility`](https://github.com/yahonda/rails/tree/per-adapter-migration-compatibility) — same dependency as #2596. Once that lands on `rails/rails#main`:

1. Switch `Gemfile` back to `rails/rails#main`
2. Mark this PR ready for review

In Phase 2 of the implicit-UNIQUE-CONSTRAINT deprecation, the global default flips from "implicit constraint + warning" (Phase 1, shipped in #2709) to "create the unique index only", matching Rails-core PostgreSQL / MySQL / SQLite. Existing migrations declared at `Migration[8.1]` or earlier opt back into the legacy implicit-constraint behavior via a `MigrationCompatibility::V8_1` module so they keep working unchanged.

## Behavior matrix

| Caller | Phase 1 (current `master`) | Phase 2 (this PR) |
|---|---|---|
| `Migration[8.2]` (Current) — `add_index unique: true` | implicit constraint + warning | **index only, no warning** |
| `Migration[8.1]` and earlier — `add_index unique: true` | implicit constraint + warning | implicit constraint + warning (preserved by V8_1) |
| `Schema.define { add_index unique: true }` | implicit constraint + warning | **index only, no warning** |
| Direct `connection.add_index unique: true` | implicit constraint + warning | **index only, no warning** |
| `OracleEnhancedAdapter.add_index_unique_creates_constraint = true` (explicit) — any caller | implicit constraint + warning | implicit constraint + warning (explicit opt-in to legacy) |
| Functional unique index — any caller | index only (no warning) | index only (no warning) — unchanged |

### Production

- `OracleEnhancedAdapter.add_index_unique_creates_constraint` default flips from `true` to `false`. Documentation updated to reflect the Phase 2 default and the explicit opt-in for legacy behavior.
- New `lib/active_record/connection_adapters/oracle_enhanced/migration_compatibility.rb`:
  - `MigrationCompatibility::V8_1` module wraps `add_index` and `create_table` so `Migration[8.1]` and earlier migrations re-enable the implicit-constraint path.
  - Thread-local key (`:__oracle_enhanced_implicit_unique_constraint`) is encapsulated as a `private_constant` of `MigrationCompatibility`, with `with_implicit_unique_constraint_enabled` / `implicit_unique_constraint_enabled?` accessors. Phase 3 deletes the whole module — key, accessors, and V8_1 — in one piece without leaving observable state behind in `Thread.current`.
- `OracleEnhancedAdapter#migration_compatibility_module_for(migration_class)` returns `MigrationCompatibility.module_for(migration_class)`, mirroring the PG / MySQL / SQLite adapters in `per-adapter-migration-compatibility`.
- `schema_statements.rb` introduces an `implicit_unique_constraint_active?` private helper that ORs the global flag with the migration-version thread-local, used by both `add_index` and `add_inline_unique_constraints`.

### Specs

- New `migration_compatibility_spec.rb` covering:
  - Migration[8.2] (current) — no implicit constraint, no warning, both `add_index` and inline `t.index` paths
  - Migration[8.1] (legacy) — implicit constraint + warning captured via `expect { ... }.to output(...).to_stderr`, both paths
  - Explicit `add_index_unique_creates_constraint = true` — overrides the Migration[8.2] default
- Mixed-strategy update of existing specs:
  - **Legacy SQL emission tests** (`should add unique constraint only to the index where it was defined`, `should emit CREATE UNIQUE INDEX and ADD CONSTRAINT for inline t.index unique: true`, `produces the same SQL whether unique index is defined inline or via explicit add_index`, the DBMS_METADATA path's `does not emit a standalone CREATE INDEX for a UNIQUE constraint's backing index`, and `emits ALTER INDEX ... INVISIBLE for an INVISIBLE constraint-backed index`) → wrapped in a new `with_implicit_unique_constraint_enabled` helper and labeled "(legacy implicit-constraint path)". These specs intentionally exercise the legacy code path; the helper makes that intent explicit.
  - **Orthogonal-feature tests** (remove_index, add_unique_constraint, schema dumper, and `structure_dump`'s `appends NOVALIDATE for foreign keys added with validate: false`) that just needed an index + constraint pair as setup are migrated to `add_unique_constraint`. Forward-compatible: Phase 3 doesn't need to touch them.
  - The `drops both the index and the same-name implicit constraint when add_index unique: true created them` spec for `remove_index` keeps the legacy `add_index unique: true` setup via the helper, because that pairing is precisely what's being tested. (Per "let Oracle handle constraint-managed indexes" — this scenario only happens for the legacy path where the index is created explicitly via `CREATE UNIQUE INDEX`, not auto-created behind a constraint.)
- `spec_helper.rb` adds `ImplicitUniqueConstraintHelper#with_implicit_unique_constraint_enabled`, included globally so any spec can opt-in to legacy.

## Out of scope

- **`remove_index` against an index that was auto-created behind an `add_unique_constraint`** still raises `ORA-01418` because Oracle's automatic cleanup drops the index together with the constraint, then `remove_index` tries to drop the (already-gone) index. Per the design principle "let Oracle handle the index it auto-creates behind a constraint", the recovery path for that case is `remove_unique_constraint :t, name: :n`, not `remove_index`. Worth a friendly error in a follow-up issue if it bites users in practice.
- **Phase 3** (delete the flag, the helper, the V8_1 module, and the implicit-constraint code path) ships one major release after Phase 2 settles. Tracked in #2702.

## Test plan

- [x] `bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/migration_compatibility_spec.rb` (5 examples)
- [x] `bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/schema_statements_spec.rb -e "implicit constraint deprecation"` (6 examples) — Phase 1 specs continue to pass; the explicit-flag-true ones still test the legacy path now
- [x] `CI=1 bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced/schema_statements_spec.rb spec/active_record/connection_adapters/oracle_enhanced/schema_dumper_spec.rb spec/active_record/connection_adapters/oracle_enhanced/structure_dump_spec.rb spec/active_record/connection_adapters/oracle_enhanced/migration_compatibility_spec.rb` — 226 examples, 0 failures, 2 pre-existing pending
- [ ] Switch `Gemfile` back to `rails/rails#main` once `per-adapter-migration-compatibility` lands
- [ ] Mark PR ready for review and run CI on full matrix

## Related

- Phase 1 (deprecation warning + opt-out flag) — #2709 (merged)
- Rails dependency — yahonda/rails `per-adapter-migration-compatibility` (same as #2596)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ group :development do
   gem "rspec"
   gem "rdoc"
   gem "rake"
-  gem "activerecord",   github: "rails/rails", branch: "main"
+  gem "activerecord",   github: "yahonda/rails", branch: "per-adapter-migration-compatibility"
   gem "ruby-plsql", github: "rsim/ruby-plsql", branch: "master"
 
   platforms :ruby do
--- a/lib/active_record/connection_adapters/oracle_enhanced/migration_compatibility.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/migration_compatibility.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module ActiveRecord
+  module ConnectionAdapters
+    module OracleEnhanced
+      module MigrationCompatibility # :nodoc: all
+        extend ActiveRecord::Migration::Compatibility::Versioned
+
+        # Thread-local key used by V8_1 to opt back into the legacy implicit-
+        # constraint behavior. Encapsulated here so that the entire
+        # MigrationCompatibility module — key included — can be removed in
+        # one piece in Phase 3 of the deprecation (#2702) without leaving
+        # observable state behind in `Thread.current`.
+        IMPLICIT_UNIQUE_CONSTRAINT_KEY = :__oracle_enhanced_implicit_unique_constraint
+        private_constant :IMPLICIT_UNIQUE_CONSTRAINT_KEY
+
+        def self.implicit_unique_constraint_enabled?
+          Thread.current[IMPLICIT_UNIQUE_CONSTRAINT_KEY] == true
+        end
+
+        def self.with_implicit_unique_constraint_enabled
+          prev = Thread.current[IMPLICIT_UNIQUE_CONSTRAINT_KEY]
+          Thread.current[IMPLICIT_UNIQUE_CONSTRAINT_KEY] = true
+          yield
+        ensure
+          Thread.current[IMPLICIT_UNIQUE_CONSTRAINT_KEY] = prev
+        end
+
+        # Phase 2 of the implicit-UNIQUE-CONSTRAINT deprecation (#2702):
+        # `Migration[8.2]+` (the current default) treats `add_index unique:
+        # true` as "create only the unique index" — matching the Rails-core
+        # PostgreSQL/MySQL/SQLite adapters. `Migration[8.1]` and earlier
+        # opt back into the pre-Phase-2 behavior of also creating a
+        # same-named UNIQUE CONSTRAINT, so existing migrations keep
+        # working unchanged. Callers that need a constraint should call
+        # `add_unique_constraint :t, :col, name: :n` directly.
+        module V8_1
+          def add_index(table_name, column_name, **options)
+            MigrationCompatibility.with_implicit_unique_constraint_enabled { super }
+          end
+
+          def create_table(table_name, **options, &block)
+            MigrationCompatibility.with_implicit_unique_constraint_enabled { super }
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb
@@ -309,7 +309,7 @@ module ActiveRecord
           execute schema_creation.accept(create_index)
 
           index = create_index.index
-          if needs_unique_constraint?(index.unique, index.columns) && OracleEnhancedAdapter.add_index_unique_creates_constraint
+          if needs_unique_constraint?(index.unique, index.columns) && implicit_unique_constraint_active?
             warn_implicit_unique_constraint_deprecation
             execute add_unique_constraint_sql(index.table, index.columns, index.name)
           end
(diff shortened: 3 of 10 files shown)


yahonda:
Rebased onto current master (e8e1677c) to pick up the enforced foreign key support (#2793), Codespaces docs (#2807, #2808), and other recent changes.

yahonda:
Superseded by #2817, which implements the add_index unique: true default flip for Migration[8.2]+ on the v4 per-adapter CompatibilityBehavior design. Closing in favor of #2817.

PR #2731, opened 2026-05-08
PoC: insert_all extras (on_duplicate skip/update via MERGE + auto-PK injection)

yahonda:
**Stacked on #2730** which is itself stacked on #2715. Marked **draft**; this PR is for **feasibility verification only**, not for direct merge.

## Goal

Demonstrate that the `out-of-scope` items called out in #2730's description are reachable with a relatively small adapter override:

| Out-of-scope item from #2730 | Status here |
|--------------------------------|-------------|
| `:skip` `on_duplicate` | ✅ via Oracle `MERGE` (caller supplies `unique_by:`) |
| `:update` `on_duplicate` | ✅ via Oracle `MERGE` `WHEN MATCHED THEN UPDATE` |
| Auto-PK via sequence injection | ✅ via `INSERT INTO ... SELECT seq.NEXTVAL, ... FROM (...)` (NOT `INSERT ALL`) |
| `returning:` option | ❌ still out of scope (Oracle MERGE / INSERT ALL don't support `RETURNING`) |

## Two Oracle gotchas this PoC documents

1. **`INSERT ALL` + `seq.NEXTVAL` is broken.** Oracle evaluates the sequence once per statement, so all rows in the multi-table insert get the same value → `ORA-00001` (unique violation). Verified empirically; the PoC routes around it by using `INSERT INTO ... SELECT ... FROM (...)`.

2. **`NEXTVAL` is forbidden inside `UNION ALL` branches** (`ORA-02287`: _sequence number not allowed here_). Oracle requires the sequence reference to live in an outer SELECT that *consumes* a UNION ALL subquery, not inside the branches themselves.

The auto-PK code path therefore looks like:

```sql
INSERT INTO "T" ("ID", "NAME", "QTY")
SELECT "T_SEQ".NEXTVAL, "NAME", "QTY"
  FROM ( SELECT 'a' AS "NAME", 1 AS "QTY" FROM dual
         UNION ALL
         SELECT 'b'        , 2          FROM dual )
```

## MERGE shape

For `on_duplicate: :skip`:

```sql
MERGE INTO "T" t
USING ( SELECT 1 AS "ID", 'a' AS "NAME" FROM dual
        UNION ALL
        SELECT 2        , 'b'           FROM dual ) s
ON (t."ID" = s."ID")
WHEN NOT MATCHED THEN INSERT ("ID", "NAME") VALUES (s."ID", s."NAME")
```

For `on_duplicate: :update`, an additional `WHEN MATCHED THEN UPDATE SET ...` is emitted. Columns in the `ON` clause are excluded from the UPDATE list (Oracle raises `ORA-38104` otherwise), with case-insensitive matching since AR's `unique_by:` typically uses lowercase symbols and `keys_including_timestamps` returns Oracle's quoted uppercase identifiers.

## Capability flags overridden

```ruby
def supports_insert_on_duplicate_skip?    = true
def supports_insert_on_duplicate_update?  = true
def supports_insert_conflict_target?      = true
```

These are set true even though the implementation requires the caller to supply `unique_by:`. With the existing AR core check that gates `unique_by:` behind `supports_insert_conflict_target?`, this is the right place to surface the requirement.

## Specs

`describe "insert_all!"` in `oracle_enhanced_adapter_spec.rb` adds three PoC examples on top of #2730's existing trio:

- Auto-fills primary keys from the table's sequence.
- Honors `:skip` `on_duplicate` via MERGE — existing rows are preserved, new rows inserted.
- Honors `:update` `on_duplicate` via MERGE WHEN MATCHED — existing rows have non-key columns updated, new rows inserted.

5/5 specs green locally.

## Known limitations / production-readiness gaps

These need addressing before this PoC could become a real PR:

1. The `split_values_list` / `split_tuple` parsers handle paren-balancing and `''` escapes, but are still string-shape-dependent. A more robust path would access the raw `insert_all.map_key_with_value` rows and re-compile per-row via Arel.
2. `returning:` is still unsupported. Implementing it would require a per-row `INSERT ... RETURNING ... INTO :bind` fallback (slow but correct), plus careful handling for the bulk case where Oracle's MERGE simply cannot return values.
3. The MERGE path requires explicit PKs (the conflict target). Auto-PK injection doesn't apply because the conflict target IS the PK by definition.
4. Composite-column `unique_by` isn't exercised in the PoC specs.
5. Column-quoting normalization between `keys_including_timestamps` (uppercase quoted via Oracle) and `unique_by.columns` (typically lowercase Ruby symbols) is handled ad-hoc with `downcase` — this might miss case-sensitive identifier scenarios.
6. No `validate_index_length!` / sequence-existence cache; sequence lookup hits the data dictionary on every call.

## Test plan

- [x] `bundle exec rspec spec/active_record/connection_adapters/oracle_enhanced_adapter_spec.rb -e "insert_all"` (5 examples)
- [ ] CI on full matrix once #2715 and #2730 merge and this rebases

This is a **feasibility check only** — the gaps above show this is not yet ready for a real merge. If the approach is acceptable, follow-up PRs can address each limitation incrementally.

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb
@@ -222,18 +222,30 @@ module ActiveRecord
         end
 
         def build_insert_sql(insert) # :nodoc:
-          if insert.skip_duplicates? || insert.update_duplicates?
-            raise NotImplementedError,
-                  "Oracle insert_all does not yet support :on_duplicate (skip/update). " \
-                  "Use a MERGE statement directly until upsert_all support lands (tracked separately)."
+          if insert.update_duplicates?
+            return build_merge_sql(insert, on_duplicate: :update)
+          elsif insert.skip_duplicates?
+            return build_merge_sql(insert, on_duplicate: :skip)
           end
 
           rows = split_values_list(insert.values_list)
-          rows_sql = rows.map { |row| "  #{insert.into} VALUES #{row}" }.join("\n")
-          # `INSERT ALL ... SELECT 1 FROM DUAL` is Oracle's bulk-insert idiom.
-          # Active Record's RETURNING expectations don't apply here -- INSERT ALL
-          # cannot carry a RETURNING clause -- so `insert.returning` is dropped.
-          "INSERT ALL\n#{rows_sql}\nSELECT 1 FROM DUAL"
+          model = insert.send(:model)
+          pk = model.primary_key
+          pk_in_keys = pk.is_a?(String) || pk.is_a?(Symbol) ?
+            insert.send(:keys_including_timestamps).include?(pk.to_s) : true
+
+          if pk_in_keys
+            # Standard INSERT ALL path. AR's `returning:` does NOT round-trip --
+            # INSERT ALL cannot carry a RETURNING clause per Oracle SQL Reference.
+            rows_sql = rows.map { |row| "  #{insert.into} VALUES #{row}" }.join("\n")
+            "INSERT ALL\n#{rows_sql}\nSELECT 1 FROM DUAL"
+          else
+            # Auto-PK injection path. INSERT ALL evaluates `seq.NEXTVAL` only
+            # once per statement (causing ORA-00001), so we route through
+            # `INSERT INTO t (...) SELECT seq.NEXTVAL, ... FROM dual UNION ALL ...`
+            # where NEXTVAL is evaluated per row of the SELECT.
+            build_insert_select_with_nextval(insert, rows)
+          end
         end
 
         # Writes LOB values from attributes for specified columns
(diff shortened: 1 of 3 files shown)


yahonda:
Closing as feasibility-only. The PoC demonstrated that:

- `:skip` / `:update` `on_duplicate` are reachable via Oracle MERGE
- Auto-PK injection works via `INSERT INTO ... SELECT seq.NEXTVAL, ... FROM (UNION ALL ...)` (sequence + trigger tables)
- Two Oracle gotchas (statement-scoped NEXTVAL in INSERT ALL → ORA-00001, NEXTVAL forbidden in UNION ALL branches → ORA-02287) and one MERGE constraint (ORA-38104 ON-clause columns can't be UPDATE'd) are documented

Plus the related discovery during review: IDENTITY columns hit the same statement-scoped issue as user-defined sequences in INSERT ALL, so even 12c+ tables would need the INSERT...SELECT path to use the auto-fill.

Not pursuing further right now — distant horizon. Keeping as a closed reference if/when this work resumes.

PR #2771, opened 2026-05-12
Mark Model.contains scope's SqlLiterals as retryable

yahonda:
## Summary

Companion to #2766. `ContextIndexClassMethods#contains` builds the `CONTAINS(col, query, label) > 0` predicate plus an `ORDER BY SCORE(label) DESC` clause from four library-controlled `SqlLiteral` / `Arel.sql` nodes:

- the quoted column inside the `CONTAINS` arguments (`SqlLiteral.new(quoted_column)`)
- the score label inside the `CONTAINS` arguments (`SqlLiteral.new(score_label.to_s)`)
- the `"0"` literal on the right of the `>` comparison
- the `SCORE(...) DESC` order (`Arel.sql("SCORE(#{score_label}) DESC")`)

All four default to `retryable: false`, and `visit_Arel_Nodes_SqlLiteral` runs `collector.retryable &&= o.retryable`, so any one of them flips the whole SELECT's `collector.retryable` to false. The CONTAINS/SCORE query is a plain idempotent CONTEXT-index SELECT, so AR core's `with_raw_connection(allow_retry: true)` should cover it.

The arguments are library-controlled — `quote_table_name` output for the column, `.to_i`-coerced score label, fixed `"0"` and `"SCORE(...) DESC"` — so marking them retryable matches the upstream Rails convention for library-generated identifiers and fixed SQL fragments: `query_methods.rb` / `predicate_builder.rb` / `finder_methods.rb` use `Arel.sql(..., retryable: true)` for the same shape, and adapter-level constants like `HIGH_PRECISION_CURRENT_TIMESTAMP` are declared the same way.

Adds a regression spec in `context_index_spec.rb` that builds the relation, walks the Arel AST through the adapter's visitor with a collector preset to `retryable = true` (matching AR's `to_sql_and_binds` preamble), and asserts the collector still reports `retryable` true after compilation.

## Test plan

- [x] `BUNDLE_ONLY=rubocop bundle exec rubocop` — clean (95 files, 0 offenses)
- [ ] CI `test` / `test_11g` — `context_index_spec.rb` exercises both `Post.contains(...).to_a` (existing scenarios) and the new collector-retryable assertion

### DIFF
--- a/lib/active_record/connection_adapters/oracle_enhanced/context_index.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/context_index.rb
@@ -339,17 +339,17 @@ module ActiveRecord
               contains_node = Arel::Nodes::NamedFunction.new(
                 "CONTAINS",
                 [
-                  Arel::Nodes::SqlLiteral.new(quoted_column),
+                  Arel::Nodes::SqlLiteral.new(quoted_column, retryable: true),
                   Arel::Nodes::BindParam.new(query),
-                  Arel::Nodes::SqlLiteral.new(score_label.to_s)
+                  Arel::Nodes::SqlLiteral.new(score_label.to_s, retryable: true)
                 ]
               )
 
               # Create comparison node: CONTAINS(...) > 0
-              condition = Arel::Nodes::GreaterThan.new(contains_node, Arel::Nodes::SqlLiteral.new("0"))
+              condition = Arel::Nodes::GreaterThan.new(contains_node, Arel::Nodes::SqlLiteral.new("0", retryable: true))
 
               # Create the where clause and order by score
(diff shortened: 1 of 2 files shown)


yahonda:
Closing — the fix does not actually achieve the stated goal.

Per CI run https://github.com/rsim/oracle-enhanced/actions/runs/25712146825/job/75494468666 the new regression spec fails because `Arel::Visitors::ToSql#visit_Arel_Nodes_NamedFunction` unconditionally sets `collector.retryable = false`. `Model.contains(...)` builds its predicate via `Arel::Nodes::NamedFunction.new("CONTAINS", ...)`, so marking the child `SqlLiteral`s as `retryable: true` is overridden the moment the visitor reaches the surrounding NamedFunction. Tracking the broader issue at #2772.

PR #2813, opened 2026-06-15
v3: Auto-enable identity primary keys in Migration[8.2]+

yahonda:
v3 redo of #2596 against the new `Rails::Migration::CompatibilityStrategy` extension point (rails/rails PR-pending, currently tracked on `yahonda/rails` branch `per-adapter-migration-compatibility-v3`).

## Behavior

`Migration[8.2]` and later default to an Oracle identity primary key when the server supports identity columns (12.1+). `Migration[8.1]` and earlier, `Schema.define`, and direct `connection.create_table` calls keep the sequence-backed default so existing schemas and dumps replay unchanged. An explicit `identity:` value always wins.

## Implementation

* New `OracleEnhanced::CompatibilityStrategy` (`extend Base::Resolver`) is wired into Rails' `compatibility_strategy_for(migration_class)` hook:
  * `V8_2#create_table` forces `options[:identity] = true` when no `:identity` is given, no sequence options are set, the connection supports identity columns, and the caller is not `ActiveRecord::Schema` (so `Schema.define` keeps the sequence default).
  * `V8_1#create_table` forces `options[:identity] = false` to maintain legacy sequence behavior on older migrations.
* `rename_table` and `drop_table` become identity-aware: they skip the paired `<table>_seq` rename/drop when the table's primary key is an Oracle identity column, so unrelated application-owned sequences with matching names are left alone.
* The schema dumper emits `identity: false` for sequence-backed primary keys on identity-capable servers, so a dump written under the new default reloads to a sequence-backed table where appropriate.

## Compared to #2596

* Same user-visible behavior at the Migration[8.2] / Migration[8.1] / `Schema.define` boundaries.
* Uses the new strategy pattern instead of the `include`-based `MigrationCompatibility` module #2596 carried. The strategy is a composition object the framework dispatches every schema operation through, so the migration class is never mutated at runtime — addressing the upstream review concern (rails/rails#57480 (comment)) that ended the v1 design.

## Tests

Existing `identity_primary_key_spec.rb` adds round-trip + identity-aware rename/drop coverage; new `migration_compatibility_spec.rb` covers Migration[8.2] / Migration[8.1] / `Schema.define` boundaries. Verified locally against Oracle Free 23ai.

## Depends on

* Rails framework PR introducing `compatibility_strategy_for` and `Migration::CompatibilityStrategy` (currently `yahonda/rails` branch `per-adapter-migration-compatibility-v3`).

The Gemfile points at that branch.


yahonda:
Superseded by #2816 (the v4 per-adapter CompatibilityBehavior implementation of the same feature). Closing in favor of #2816.

PR #2814, opened 2026-06-15
v3: Phase 2 — flip add_index unique: true default in Migration[8.2]+

yahonda:
v3 redo of #2710 (Phase 2 of the implicit-UNIQUE-CONSTRAINT deprecation #2702) against the new `Rails::Migration::CompatibilityStrategy` extension point.

## Behavior

`Migration[8.2]` and later treat `add_index :col, unique: true` as "create only the unique index" — matching Rails-core PostgreSQL/MySQL/SQLite. `Migration[8.1]` and earlier keep the legacy behavior of also creating a same-named UNIQUE CONSTRAINT, so existing migrations replay unchanged. Callers that need a constraint on Migration[8.2]+ should call `add_unique_constraint :t, :col, name: :n` directly.

## Implementation

* `OracleEnhanced::CompatibilityStrategy::V8_1` now sets a private `_implicit_unique_constraint: true` flag in `create_table` and `add_index` (in addition to the identity-default backstop from the previous PR). The adapter consumes and deletes the flag before further processing, so it never reaches DDL emission.
* New `implicit_unique_constraint_active?` adapter helper ORs the global `add_index_unique_creates_constraint` flag with the per-call `_implicit_unique_constraint` flag.
* `OracleEnhancedAdapter.add_index_unique_creates_constraint` default flips from `true` to `false`. The flag remains as an explicit opt-in for projects that want the legacy behavior project-wide.
* Specs that exercise the legacy path now use `with_implicit_unique_constraint_enabled` in `spec_helper.rb`, which toggles the global flag around a block.
* `migration_compatibility_spec.rb` gains a second top-level describe block covering both Migration[8.2] (no implicit constraint, no warning) and Migration[8.1] (implicit constraint + deprecation warning), inline `t.index` cases, and an explicit-flag override case.

## Compared to #2710 Phase 2

* Same user-visible behavior at the Migration[8.2] / Migration[8.1] boundary.
* Uses the v3 strategy pattern instead of `include`-based modules. Notably, this PR uses an **options-flag** signal (`_implicit_unique_constraint`) rather than the thread-local `with_implicit_unique_constraint_enabled` carried by #2710 Phase 2. Migrations are single-threaded, the strategy is the wrapper itself, and the flag is consumed/deleted by the adapter before DDL emission — so thread-local state is not needed and the explicit options flag keeps the data flow visible.

## Tests

All exercised against Oracle Free 23ai: `migration_compatibility_spec.rb` (combined identity + unique), `identity_primary_key_spec.rb`, `schema_dumper_spec.rb`, `schema_statements_spec.rb`, `structure_dump_spec.rb` all pass.

## Depends on

* **#TBD: Auto-enable identity primary keys in Migration[8.2]+** — this PR builds on that branch (the strategy plumbing and `V8_1#create_table` identity backstop). Merge that first.
* Rails framework PR introducing `compatibility_strategy_for` and `Migration::CompatibilityStrategy` (currently `yahonda/rails` branch `per-adapter-migration-compatibility-v3`).


yahonda:
Superseded by #2817 (the v4 per-adapter CompatibilityBehavior implementation of the same feature). Closing in favor of #2817.

PR #2816, opened 2026-06-28
Default to identity primary keys in Migration[8.2]+

yahonda:
Adopt the standalone `ActiveRecord::Migration::CompatibilityBehavior` extension point (rails/rails#58162, tracked on `yahonda/rails` branch `per-adapter-migration-compatibility-v7`) to default new-enough migrations to Oracle identity primary keys. Supersedes #2813 and #2596.

## Status (2026-09-03)

* Rebased onto current master (after #2852 restored the JRuby rows and moved the pin to rails/rails@74bb87c5), on top of #2823's rebased commit; the branch content is unchanged.
* rails/rails#58162 is open and rebased onto current rails main.
* Full spec suite green locally against Oracle Free (1148 examples, 0 failures); CI green.

## Behavior

A migration declaring `8.2` now creates an identity primary key on Oracle Database 12.1 or higher, with no change to the migration code:

```ruby
class CreatePosts < ActiveRecord::Migration[8.2]
  def change
    create_table :posts do |t|   # id: NUMBER GENERATED BY DEFAULT AS IDENTITY
      t.string :title
    end
  end
end
```

No `posts_seq` sequence is created, and inserts no longer query `posts_seq.NEXTVAL` — the database assigns the id.

The same table under `Migration[8.1]` or earlier keeps the sequence-backed primary key, so every existing migration replays unchanged:

```ruby
class CreatePosts < ActiveRecord::Migration[8.1]
  def change
    create_table :posts do |t|   # id: NUMBER + posts_seq sequence, as today
      t.string :title
    end
  end
end
```

An explicit `identity:` always wins over the version default, in both directions:

```ruby
class CreatePosts < ActiveRecord::Migration[8.1]
  def change
    create_table :posts, identity: true do |t|   # identity, even on 8.1
      t.string :title
    end
  end
end

class CreateAudits < ActiveRecord::Migration[8.2]
  def change
    create_table :audits, identity: false do |t| # sequence, even on 8.2
      t.string :action
    end
  end
end
```

### Resulting primary key by caller and option (Oracle Database 12.1+)

| caller | (no `identity:`) | `identity: true` | `identity: false` |
| --- | --- | --- | --- |
| `Migration[8.2]` and later | **identity** (the only cell this PR changes) | identity | sequence |
| `Migration[8.1]` and earlier | sequence | identity | sequence |
| `Schema.define` / `ActiveRecord::Schema[x.y].define` / direct `connection.create_table` | sequence | identity | sequence |

On Oracle Database releases before 12.1 every cell resolves to a sequence-backed key, except `identity: true`, which raises `ArgumentError` — both as today. A `Migration[8.2]`+ `create_table` passing a sequence-backed option (`sequence_name:`, `sequence_start_value:`, `primary_key_trigger:`) behaves like the middle row: the option opts it out of the identity default instead of raising about an `identity: true` the author never wrote.

Schema dumps written by older releases reload unchanged (schema loads keep the sequence default), and identity-backed tables are dumped with `identity: true`, so dump -> load round-trips preserve both kinds.

## Implementation

* Fills in the `V8_2`/`V8_1` behaviors that #2823 introduces as empty skeletons:
  * `V8_2#create_table` forces `options[:identity] = true` when no `:identity` is given, no sequence options are set, the connection supports identity columns, and the caller is not a schema load (checked via `ActiveRecord::Schema::Definition`, which both plain `Schema.define` and versioned `ActiveRecord::Schema[x.y].define` include), so schema dumps keep the sequence default.
  * `V8_1#create_table` forces `options[:identity] = false` to keep the pre-8.2 sequence default on older migrations.
* `rename_table` and `drop_table` become identity-aware: they skip the paired `<table>_seq` rename/drop when the table's primary key is an Oracle identity column, so unrelated application-owned sequences with matching names are left alone.
* The schema dumper emits `identity: true` for identity-backed primary keys. Sequence-backed primary keys carry no annotation: `identity: false` is the adapter's baseline default (`Schema.define`, direct `connection.create_table`, and Migration[8.1] and earlier all keep the sequence default), so emitting it on every sequence-backed table would be redundant for reload and add schema.rb churn.

## Tests

`identity_primary_key_spec.rb` (round-trip + identity-aware rename/drop; the previously always-skipped “identity counterparts” block now runs) and `migration_compatibility_spec.rb` covering the version matrix: Migration[8.2] / Migration[8.1] / Migration[7.0], plain `Schema.define` and versioned `ActiveRecord::Schema[8.2].define` loads, explicit `identity:` overrides, and the sequence-option / `primary_key_trigger:` opt-outs. CI runs the suite against Oracle.

## Depends on

* **#2823: Add CompatibilityBehavior plumbing for per-adapter migration compatibility** — merge that first; its commit is the first of the two commits in this PR's diff.
* rails/rails#58162, which introduces `compatibility_behavior_for` and `Migration::CompatibilityBehavior` (head: `yahonda/rails` branch `per-adapter-migration-compatibility-v7`). #2823's commit points the Gemfile at that branch.

Independent of #2817: after #2823, either PR can merge first; whichever merges second needs a trivial rebase (both touch `V8_1#create_table` and append to `migration_compatibility_spec.rb`).

## Review follow-up (2026-07-02)

A deeper review pass added the following fixes, since folded into the single commit:

* **Versioned schema loads kept the sequence default only for the unversioned form.** `rake db:schema:dump` writes `ActiveRecord::Schema[8.2].define`, whose anonymous class is not a subclass of `ActiveRecord::Schema`, so the `is_a?(ActiveRecord::Schema)` guard missed it and `db:schema:load` would have recreated every sequence-backed table as identity (breaking dump -> load -> dump idempotence). The guard now checks `ActiveRecord::Schema::Definition`, with a spec loading through `ActiveRecord::Schema[8.2].define`.
* **`primary_key_trigger: true` now opts out of the identity default.** Previously a `Migration[8.2]` `create_table :t, primary_key_trigger: true` raised about combining it with an `identity: true` the author never wrote.
* **The "identity counterparts" spec block was always skipped.** Its guard read `@oracle12c_or_higher`, which is never assigned in that file; it now uses the `database_version` comparison. The unskipped NEXTVAL assertion was tightened to `/\.NEXTVAL/` so it does not match the `has_primary_key_trigger?` lookup SQL, which contains the literal `NEXTVAL INTO :NEW`.
* Spec/comment cleanups: non-deprecated `database_version` comparison in a `before(:all)` hook, removal of an unused table cleanup.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,9 +7,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 group :development do
   gem "rspec"
   gem "rake"
-  # rails/rails@74bb87c5 = merge of rails/rails#58684 (load Active Support on
-  # JRuby without Ractor); bump per follow-up Rails change
-  gem "activerecord",   github: "rails/rails", ref: "74bb87c5888b3489fc519f048e6a0cd9cc615a5d"
+  gem "activerecord",   github: "yahonda/rails", branch: "per-adapter-migration-compatibility-v7"
   gem "ruby-plsql", github: "rsim/ruby-plsql", branch: "master"
 
   platforms :ruby do
--- a/lib/active_record/connection_adapters/oracle_enhanced/compatibility_behavior.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/compatibility_behavior.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module ActiveRecord
+  module ConnectionAdapters
+    module OracleEnhanced
+      module CompatibilityBehavior # :nodoc: all
+        Base = ActiveRecord::Migration::CompatibilityBehavior
+        extend Base::Resolver
+
+        # A behavior applies to migrations declaring its own version and older:
+        # Migration[8.2]+ resolves to V8_2, Migration[8.1] and earlier to V8_1.
+        class V8_2 < Base; end
+        class V8_1 < V8_2; end
+      end
+    end
+  end
+end
--- a/lib/active_record/connection_adapters/oracle_enhanced/compatibility_behavior.rb
+++ b/lib/active_record/connection_adapters/oracle_enhanced/compatibility_behavior.rb
@@ -9,8 +9,40 @@ module ActiveRecord
 
         # A behavior applies to migrations declaring its own version and older:
         # Migration[8.2]+ resolves to V8_2, Migration[8.1] and earlier to V8_1.
-        class V8_2 < Base; end
-        class V8_1 < V8_2; end
+
+        # Migration[8.2]+ migrations default to an Oracle identity primary
+        # key when the server supports it. Schema loads (plain and versioned
+        # `ActiveRecord::Schema[x.y].define`, both including
+        # +ActiveRecord::Schema::Definition+) and direct adapter calls keep
+        # the adapter's pre-existing sequence default so that schema dumps
+        # written by older releases reload unchanged. An explicit
+        # `identity:` value always wins.
+        class V8_2 < Base
+          def create_table(table_name, **options)
+            options[:identity] = true if default_to_identity?(options)
+            super
+          end
+
+          private
+            def default_to_identity?(options)
+              return false if migration.is_a?(ActiveRecord::Schema::Definition)
+              return false if options.key?(:identity)
+              return false if options[:sequence_name] || options[:sequence_start_value]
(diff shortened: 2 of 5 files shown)


yahonda:
Builds on #2823 (its commit is included here), which in turn depends on rails/rails#58162 — the Active Record change that adds the compatibility_behavior_for extension point. CI here runs against that rails branch until #58162 merges.

PR #2818, opened 2026-06-28
Pass the name: keyword to Arel::Table.new in specs

yahonda:
rails/rails#57829 (merged 2026-06-24) refactored `Arel::Table.new` to take the table name as a `name:` keyword argument instead of a positional one:

```ruby
def initialize(name: nil, as: nil, klass: nil, type_caster: klass&.type_caster)
```

The Gemfile already targets rails/rails `main`, so the existing `Arel::Table.new(:users)` spec setups now raise `ArgumentError: wrong number of arguments (given 1, expected 0)`, and the Arel / `columns_for_distinct` specs fail on every Ruby (CRuby and JRuby).

This passes the table name as `name:` in the affected specs.

## On equivalent tests

rails/rails#57829 itself added no new tests — every test change in it is a balanced, mechanical migration of `Arel::Table.new(...)` call sites (lib + all Arel test cases) to the `name:` keyword. This change is the Oracle adapter's equivalent migration for its own Arel specs; there is no new behavior in rails/rails#57829 that would warrant additional test cases here.


yahonda:
JRuby failures will be taken care of separately.

